PR #41, opened 2016-08-17
SWIK-244 user profile

rmeissn:
### Do users already exist?
Yes! test and test2 are 2 users, that I've created. test is full user profile, where as test2 is a minimal profile.

### How to reach a user profile?
Go to /user/test or /user/test2 or whatever username you like.

### How to login?
You can login on both through passwords: 123456aA

Happy testing ;)


ali1k:
- can you use `link` css class for the tab items.
- can we have 'username' or 'email address' option (similar to Github) in the login modal? I am not sure if it belongs to this branch or goes to another JIRA ticket. Which email address should I use to login now and check the functionality?

rmeissn:
- You want to use 'link' css class because the mouse doesn't change to a hand on the tabs? I tried to add it, but it doesn't change to a hand with the link class or changing the whole item to '<a/>'
- We agreed at the hackathon to login through email and password instead of username and password. See updated description for available emails

ali1k:
with regards to the hand issue, I added the `link` class to the items and now the cursor has changed properly.

ali1k:
I changed the first name & last name but cannot see the result. Can you show it above the username (similar to Github)?
How can I got to edit profile when I am in the profile page?

rmeissn:
Hasn't changes in my browser....but sure. If you think it's better.
The UI code is already included, but the backend doesn't deliver fore- and surname for public uses. I've already requested this as a feature request at Kurt ;)

I don't understand the last question. Whenever you are on your own profile and you are logged in, you are in the edit view.

ali1k:
what browser are you using? I checked with latest version of Chrome, Firefox, Safari and it shows the hand cursor now!
For first-last name, would be good to have the placeholder there and as soon as you get the data from service, fill the data in!
With regards to the profile edit, I think the issue is that the browser doesn't preserver the user data, so when I go to profile page it is not in the edit mode. I am not sure if it should always show the edit mode when I go to the profile, there should be an edit button to go to the profile edit mode.

rmeissn:
Worked after npm install....like in the previous comment.

It's all there....nothing to add here, because it's implemented ;)

Yeah, that's the lost state bug. As soon as this one is fixed, it'll work as expected.
In case you're logged in and you're on YOUR profile page (e.g. /users/ali1k), you see the detailed stuff and are able to change your information. Your not in edit mode on other profiles.
As you will see a much more detailed version (that is styled differently) of your user profile, I don't think you need to look at your public profile that often....You won't even arrive on settings (like now). You will arrive on "My Decks" that is much more useful.

ali1k:
ok, good! I still think I should be able to see my profile page as public see when I am logged in! we can change it later... I am going to merge it to master... thanks

PR #42, opened 2016-08-18
Swik 283 user registration

dpaun:
The user registration component is available through the login dialog or at http://localhost:3000/signup.
Checking availability of username and email is performed before the form is submitted and, if needed, some suggestions for alternative username are provided.


kadevgraaf:
One  small thing is that when you directly click on the link “Don't have an account? Sign up here.” in the signup-modal, then it shows the registration form, but after 1 second it moves back to home page. 
When you click right or left of the link, in blank space, then you also navigate to the registration form, but correctly stay on it Perhaps a button would be more intuitive?

The maximum length of e.g. a username, can be very long. Perhaps should we set and check for a maximum length?
This is related to an error I encountered when filling in username "_<script type="text/javascript">     alert("Hello!"); </script><script type="text/javascript">     alert("Hello!"); </script>_...etc" - in this case the Sign-Up button does not work/give any feedback.
In case of above error with very long username you can press the signup-button twice (or more). The CAPTCHA status is then also not saved (internally?) in this case - the check "I am not a robot" remains, but an overlay error is given: "Error! Wrong captcha. Please verify that you are human.". Perhaps this is not a problem anymore when above check is implemented.

When I filled in "kadevgraaf" as username (instead of the long username above) the form did work well and I got a very good-looking overlay message:
"Thanks for signing up! Thank you. You have successfully registered. Please sign in with your new credentials."
After which I can login, log out, see user profile/deck-page, etc.. 👍 

I will take a quick look at the code as well.

dpaun:
Thanks for the comments!
The default behavior of the signup link, which caused redirecting to home page, is now prevented.
Username validation is improved to prevent long (>64 characters, as in Google) usernames and to ensure that it contains only alphanumeric characters, which is what the microservice requests.

kadevgraaf:
Looks good!

kadevgraaf:
Shall we leave this branch open? Near future work on email notification might still affect part of this task.

dpaun:
I'm not sure. If we leave the task 283 in Jira open, then we can also leave the branch open.

kadevgraaf:
Ah. confirm email is another task = SWIK-285. Then closing this branch and SWIK-283 seems best.

PR #43, opened 2016-08-18
SWIK-242 Docker build based on another base image

rmeissn:
Outsourced static commands like install git/bower/webpack, etc. in order to speed up build process.


rmeissn:
Seems to work like expected.
I've created an automated build of the base image on Dockerhub, so it will get updated with every new commit to the https://github.com/slidewiki/platform-runtime repository.

PR #44, opened 2016-08-18
SWIK-379 send uploaded PPTX in front-end to import-microservice to be read by PPTX2HTML filereader

TBoonX:
Changed code for Import component, actions and store. **No** new components or changed routes.
Will be reused for SWIK-366.


rmeissn:
Some errors persist when executing some uploads. Please fix these first.

PR #46, opened 2016-08-22
Swik 451 solving problems with comments and sources tabs

dpaun:
Comment and Sources tabs are working again - displaying mock-up data. Currently, most of the slides do not have any comments. 'Introduction' slide contains mock-up comments plus some created using the form.


dpaun:
In short: Comments and Sources tabs stopped working properly due to recent deck/slide id changes. This branch solves this problem.

kadevgraaf:
Looks good!
The amount of comments is not yet shown, however, I recall this was never the case.

PR #47, opened 2016-08-23
SWIK-366 enhance import page

TBoonX:
I added a loader action. I added code in services/deck.js. I deleted the import route.
ATTENTION: I merged from master and now every page returns "Cannot GET /". I cant fix it.


TBoonX:
I checked the code out in another directory and there it is working. Could be tested!

kadevgraaf:
Looks good. Nice progress bar!

If the import-service is not accessible/down then the upload modal is not working (remains overlay over system after submit button)

If there are missing path variables in configs/microservices.js (protocol, host, path), and the request to the import-services is sent to a non-existing (404) page (so the deck-service does nothing), then still the import page gives a success message ("slides uploaded"). This is technically correct: the progress bar is about uploading, not a successful import (right?). 

Minor detail: after uploading the filename of the *.PPTX file is shortened
screen shot 2016-08-29 at 17 01 44

On https://slidewiki.atlassian.net/wiki/display/SWIK/Add+a+deck fields license and Theme have the following specification:
Field:       Required?                                                                                                     Notes:
License	Will be set to a default CC licence                                                             Need to agree
Theme	Will be set to default theme unless user selects a different theme	 

Currently the default license and theme is not set. License is compulsory ("need to agree").  

Because the default theme is not set by default :) in the form, there is an error, which remains even after correction (setting the default theme):
400
Type: Bad Request
Description: child "description" fails because ["description" is not allowed to be empty]
Action required: Restart your action. 
I think this can be easily resolved by removing option "Select Theme" in the form. 

When leaving the description field empty you get error:
400
Type: Bad Request
Description: child "description" fails because ["description" is not allowed to be empty]
Action required: Restart your action.
However, according to the checks in the form and (seemingly) https://slidewiki.atlassian.net/wiki/display/SWIK/Add+a+deck the description field can be empty.

PR #48, opened 2016-08-24
SWIK-454 header user dropdown

rmeissn:
Created a dropdown that is only visible for logged in users to access the user profile and some settings. Additionally, I've modified the header and footer bar to better match their purposes.

Styling of the notifications popup and the user dropdown do not match. This will require some time to fix, as they do have a complete different implementation approach. The notification approach hasn't worked for the user dropdown and violates the page/container boundaries (may cause problems in future times).
Please think about these issues and create Jira tasks, if necessary.


kadevgraaf:
Looks good! 
Some small comments:
- When showing the menu for the first time it is a bit smaller.
- Link to "my profile" and "my settings" in the menu leads to the same page (will these be 2 separate pages?).
- In firefox dev 49 the links in the menu cannot be opened as new tab by clicking with, e.g., middle mouse button or "open in new tab". Perhaps this is solved using semantic UI by adding a class "anchor" or "clickable" (I do not know the exact keyword, but this might work; just a sugestion :) )
- After signing out you see you public profile. This is fine I guess. An alternative would be to to the home page.  Perhaps should we create a JIRA task to have a link to the public profile of a user in the drop-down menu or "my profile" page of a user?

rmeissn:
Thanks for the review!

About your listing:

1. I'll investigate into this one
2. As I've discussed with Ali, it would be better to separate public and private user profiles to different URLs. As of this work, these two links will link different views.
3. I'll investigate into this one
4. You'll only "redirected" to the public profile in case you were on your private profile. Otherwise you'll stay at the same page - UserProfile component listens for the login state. As I do not want to always leave the page I'm on by logging out, the redirect would be a custom implementation for specific components/pages. As of the current component state, you can't access your public profile when you're logged in. This may change in future iterations, but having a link to the public profile would be nice. Maybe this will result in one task "Separate public and private profile and link to these"

kadevgraaf:
For point 1: maybe this was in my browser cache - I cannot reproduce it myself :)

rmeissn:
Please review again!

About your listing:

1. Not reproducible: ignored
2. Implemented separate URLs for public and private user profiles. All reachable by the head dropdow.
3. Implemented (for header dropdown and private users category box)
4. By logging out on a private page, you're now redirected to the home page. Logging out on all the other pages won't redirect.

kadevgraaf:
Looks good!

PR #49, opened 2016-08-29
SWIK-455 save user data

TBoonX:
I added a plugins folder to the project. There is one directory for each plugin. I added it to the application and used it in login and logout.
Also I added the res variable to the fluxible context.


rmeissn:
Seems to work fine, except for the described issue in Jira. Good to merge

PR #52, opened 2016-09-02
SWIK-481 connect revisions panel to deck service

stavmars:
Connected content history panel in order to get a deck or slide's revisions. In order to find which revision is active, the specific node's immediate parent (with revision id) is extracted from the spath and it is requested from the deck service. So two parallel requests are made to the deck service (one for the parent deck and one for the selected item's revision history). This branch also addresses jira ticket SWIK-406.


kadevgraaf:
Looks good!
Some comments, which are probably caused by other parts of the system not yet implemented:
- The slide revisions are not yet shown. The user is also not updated when revisioning (maybe this is due to the status of slide revisioning + connecting it to slide-content-edit. Please ask Marios about the status of this and then doublecheck if it works in the history tab. Also revert should be possible for slides but I cannot test because it only shows 1 revision always for each slide). The time of the last edit is updated and shown accurately!
-there is a revert button, but it does not work. Perhaps it is best not to show the button until we are sure it works in the future.
- If you navigate to an earlier revision, e.g. Revision 1 of deck 1, then still all the links in the decktree point to the latest revision you viewed, e.g. revision 2. the title is not updated. After refresh the links are correctly updated though - this is probably due to the decktree-store not being updated imediately. Perhaps when you click on a revision in the history tab then it should act like when you refresh a page. So a "new page"/hyperlink instead of an internal navigate action. I will make a ticket for this. 

I have merged this into the master. Please do not delete the branch but continue working on it. Next work could be user notifications when a revision is done. Also, the history tab can show the number of revisions on the tab (like now is done for Questions and sources tab).

PR #57, opened 2016-09-12
Swik 498 using real user id for comments

dpaun:
The discussion-service and front-end components are modified to use the logged-in user id (from UserProfileStore) when creating comments and replies. If the user is not logged-in, he/she can not create comments.
Also, the comment count (received from the micro-service) is shown, like on 'Questions' and 'Sources' tabs.


kadevgraaf:
Looks good! Both in firefox and Chrome.

Minor detail: the comment count is only updated after refreshing the deck of slide (e.g., navigating to different slide/deck and then back)

PR #59, opened 2016-09-13
Swik 505 pptx import feedback

dpaun:
This pull request addresses **SWIK-503** (Cancelling upload of PPTX) and **SWIK-505** (add maximum pptx file size + feedback after clicking on "Add deck" button for import).
- Upload modal window is hidden to prevent the problem with multiple modal windows staying open when 'Add Deck' is executed more than once.
- Max file size (currently 300M) is displayed to the user.
- Import progress bar is improved - during the pptx import, the deck-service is queried for the current number of slides in the newly created deck.


kadevgraaf:
Cool progress bar! Like the youtube uploader. I think this is really what Ali and I envisioned. 
One minor detail: When uploading a 2 slide presentation, in the end the upload status says: "Imported 5 of 2 slides". In another test attemt the final message is "slides uploaded!", so correct.
- I tested by uploading a pptx file of 400 mb (not an actual PPTX file, but a *.dmg file renamed to *.pptx). This works -> upload button remains disabled in the modal.
- Button "add deck" is disabled until slide is uploaded and processed.
- Upload window is hidden once file is being uploaded (after clicking on "upload" button)
Tested in Chrome.

PR #60, opened 2016-09-13
Swik 388 slide view display

huwf:
This branch has SWIK-340 and SWIK-388.  Both are in QA


kadevgraaf:
Looks good.
The theme is loaded (not yet dynamically, but for that to work we need to store themes first) in slide content view and editor.

PR #62, opened 2016-09-14
SWIK-525 added configuration of microservices when container starts

bwulff:
SWIK-525

The addresses of the microservices are taken from a set of environment variables and filled into a template, the processed template is written to /nodeApp/configs/microservices.js.
Defaults values for the env variables are defined in the Dockerfile. They can be overwritten via -e argument to the docker command, the environment directive in compose files or an .env file.


bwulff:
Klaas [1:55 PM]  
hmmm, 
"docker: Error response from daemon: oci runtime error: exec: "./entrypoint.sh": permission denied.”
Even with SUDO

bwulff:
probably entrypoint script has wrong mode...

bwulff:
made entrypoint script executable with commit cbc5073

kadevgraaf:
ERROR in main.js from UglifyJs
Unexpected character '`' [./~/identicons/index.js:38,0]

kadevgraaf:
I'll merge, let see what happens.

PR #79, opened 2016-09-20
Swik 540

stavmars:
This PR addresses the following tickets:
SWIK-531 Save function in slide content editor not working
SWIK-540 Slide update in services/slide.js returns dummy data
It also includes changes to some deck tree related actions, in order to get the logged-in user's id.


kadevgraaf:
Works well! 
Revision of slide + user shows up perfectly in history tab.
Tested in firefox and Chrome.
If you are not logged in, slide save does not work, but this is ok -> in a different task (https://slidewiki.atlassian.net/browse/SWIK-588) we should redirect/disable slide edit when you are not logged in.

PR #83, opened 2016-09-21
SWIK-493_refresh_page_after_logout

TBoonX:
Added reloading of page when user is deleted (in plugin userContext)


kadevgraaf:
Looks good. Works on homepage, user profile page, deck view page in chrome and firefox. Thanks!

PR #85, opened 2016-09-21
Swik 572 update activity feed

dpaun:
New activity is now added to the activity feed when a new comment/reply is added.


kadevgraaf:
Works well in firefox and Chrome. Very clear code addition in ActivityFeedStore.

PR #87, opened 2016-09-22
SWIK-454 improve header dropdown

rmeissn:
Improved the dropdown by some suggestions of Abi. Still not perfect, but a little better.


TBoonX:
{color:red}The username is uppercase{color}
Sign out does no navigation - heading to homepage?
My Stats are not available through dropdown.

abijames:
Definitely an improvement and sorted out many of the issues! I believe the outstanding a11y issues are semantic-ui now. @kadevgraaf can you QA before merging?

PR #88, opened 2016-09-23
Swik 573 show activities for selected slide/deck

dpaun:
Activity feed now shows activities for the selected slide/deck.
If a deck is selected, activities for its subdecks and slides are also shown.
Most of the code changes to enable this were done on the activity-service.
This branch also solves problems with slide/deck links (SWIK-596).

### DIFF
--- a/components/Deck/ActivityFeedPanel/ActivityItem.js
+++ b/components/Deck/ActivityFeedPanel/ActivityItem.js
@@ -22,7 +22,7 @@ class ActivityItem extends React.Component {
             fontStyle: 'italic',
             fontWeight: 400
         };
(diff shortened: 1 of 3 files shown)


ali1k:
it looks good. Can we just have a minor addition to show a text in case there is no activity for the selected item?
For instance to show `There is yet no activity around this content'
and later on to guide user to create an activity e.g. be the first one to comment on this slide!

dpaun:
Hmm... I have added something like this. When I test this branch on my localhost it displays "There are currently no activities for this slide." (or deck) when there are no activities.

ali1k:
ooh right! I see it now after pull! will merge it now...

PR #113, opened 2016-09-27
Swik 575

stavmars:
This pull request adds funtionality to the fork button. It specifically implements the fork functionality of the old platform and creates a new deck revision with the same data as the currently selected one.


ali1k:
we still need to add a notification message to inform user about the new revision. The language field in DeckEdit does not work. When adding a subdeck, it should automatically add the language

PR #117, opened 2016-09-28
Swik-514 semanticui modal

TBoonX:
Replaced react modal with semantic-ui modal
It is not accessible but there is no more overlapping anymore
I also made it simpler to toggle the login modal

### DIFF
--- a/actions/loadAddDeck.js
+++ b/actions/loadAddDeck.js
@@ -9,7 +9,7 @@ export default function loadAddDeck(context, payload, done) {
         context.executeAction(navigateAction, {
             url: '/'
         });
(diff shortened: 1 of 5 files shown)


ali1k:
It looks fine. For the first release, we should make sure that the issue with overlapping is resolved. I think we can use semantic-ui modal for now and later on try to make it more accessible. @abijames is it OK to use the semantic-ui modal for now? if we cannot quickly fix the overlapping issue which is very critical now?

abijames:
@ali1k I suggest merging this branch for now. I will test it tonight to see the impact for both mobile use and accessibility. I agree we need to sort the overlapping bug out before the release but we also can't leave sign in to be inaccessible for long.

PR #119, opened 2016-09-28
Swik 612 handle some service unavailable errors

dpaun:
This branch handles some errors related to unavailability of some services (discussion, activity, notifications).


ali1k:
we still need to adapt this change based on the new the error handling actions and then will be able to merge

PR #126, opened 2016-09-28
Reveal fork 2

huwf:
Should make the view and edit mode match the presentation mode and look nice.  Other major thing to note is that we are now using a fork of reveal.js and have added that as a submodule.  Because that caused thousands of errors, I've add an .eslintignore file so we don't have to see them.

It's got a few minor changes to the reveal.css code to take that away from affecting body, and a couple of minor changes to the HTML of the slide view and edit components


kadevgraaf:
Looks good! 👍 
Some feedback:
preferably next time link it to a SWIK number :)
Some fonts become too large, e.g.; http://platform.manfredfris.ch/Presentation/478-1/#/slide-2580-1
Some content is shown outside slides, e.g.; http://platform.manfredfris.ch/Presentation/478-1/#/slide-2596-1 We might fix this by changing the surroundings of slide content in reveal.js (background?) or putting pptx2html in a container that has style (or giving ppt2html div style) in which overflow/scrolling/outside content is disabled.

huwf:
Yeah, I plan to add some tweaks to the `white.css` theme tomorrow now that we've merged this.  I think if we make the default font size smaller for a few things then those problems should go away.  We'll have to discuss how that relates to what is being imported.

View mode should have `overflow:hidden` so that you will know how it will look, whilst edit should allow overflow so you can see what you need to change.

And yes, next time I'll relate it to a SWIK number! :smile:

kadevgraaf:
Sounds good!
I also noticed my precious resize and drag buttons in the slide content editor are broke, which might have to do with the new reveal.js. Nothing we cant fix. Do you have an idea what might cause it?
One other thing I noticed is that the cursor is now 'text-select' by default in the slide content editor.
CKeditor seems to work better now with its positioning and with the boldface, italic, etc.. which solves https://slidewiki.atlassian.net/browse/SWIK-646

PR #132, opened 2016-09-29
Swik-640 same modals

TBoonX:
Just included the js and css files for sweetalert2


ali1k:
see documentation at https://github.com/slidewiki/slidewiki-platform/wiki/Modals-and-notifications

PR #139, opened 2016-09-29
Swik 644 display notifications for user's decks slides

dpaun:
Page /notifications now displays notifications about user's decks and slides. Filtering, which was previously on the left side (by subscribed content, by type of activity), is removed because we currently have only one type of notifications - comments about user's decks and slides.
If there are no notifications, there is a message displayed.


kadevgraaf:
We hope to merge this after the release.

bwulff:
Tested by @dpaun : Not yet in shape for merging. Closing it.

PR #146, opened 2016-09-30
SWIK-658 Prevent continuous user querying

rmeissn:
I have to assist to test it, because I need to delete a user in the database, after someone has logged in.


kadevgraaf:
Tested in Firefox -> register user prevent + sign-in + delete in database + refresh page == logout and no continuous errors anymore.

PR #148, opened 2016-09-30
PDF generated via download icon (needs pdfservice updated to equivale…

kadevgraaf:
…nt branch)

### DIFF
--- a/components/Deck/ContentPanel/ContentActions/ContentActionsFooter.js
+++ b/components/Deck/ContentPanel/ContentActions/ContentActionsFooter.js
@@ -5,6 +5,7 @@ import {connectToStores} from 'fluxible-addons-react';
 import SlideControl from '../SlideModes/SlideControl';
 import expandContentPanel from '../../../../actions/deckpagelayout/expandContentPanel';
 import restoreDeckPageLayout from '../../../../actions/deckpagelayout/restoreDeckPageLayout';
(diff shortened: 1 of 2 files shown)


kadevgraaf:
@Sole-Valero do you wan to record your QA/testing findings and/or screenshots in a new task on JIRA?

Sole-Valero:
Calling the pdf-service to export deck 8, we obtain a pdf, but the content of the exported slides is different from the original ones. It seems like if it is using different font styles, so the layout of  content is not correct. I attacht different images to show the result. 
slide8_2
slidepdf8_2


Also, if the pdf-service is called from a particular slide, it doesnt´t work, answering with error:
{"statusCode":400,"error":"Bad Request","message":"child \"id\" fails because [\"id\" with value \"8-17\" fails to match the required pattern: /^[0-9]+$/]","validation":{"source":"params","keys":["id"]}}

Sole-Valero:
About the error from a particular slide, the request made from the button is incorrect. Now it calls to:
`http://pdfservice.experimental.slidewiki.org/exportPDF/8-17`

But it should call it as:
`http://pdfservice.experimental.slidewiki.org/exportPDF/8?slides=17`

That is:
`ttp://pdfservice.experimental.slidewiki.org/exportPDF/{deckId}?slides=numberSlide`

PR #150, opened 2016-10-03
SWIK-642 Added notifications for empty comment fields

dpaun:
This is a small fix that displays an error when comment title or text are empty.
In the future, we will consider allowing comment title to be empty.


dpaun:
I see that this PR was merged and then reverted. Was there a problem with it?

kadevgraaf:
Hi Dejan,
I am not sure, since I cannot find any mention here or Slack and JIRA of the revert reason (sorry about that), but perhaps it was reverted just to be sure this was not the cause.
We can try to merge again and see how it works out.

PR #151, opened 2016-10-03
Swik 673 update signup form

abijames:
updated sign up form to improve messaging, accessibility & add organisation field.

### DIFF
--- a/components/User/UserRegistration/UserRegistration.js
+++ b/components/User/UserRegistration/UserRegistration.js
@@ -30,6 +30,13 @@ class UserRegistration extends React.Component {
                         prompt: 'Please enter your last name'
                     }]
                 },
+                organisation: {
+                    identifier: 'organisation',
+                    rules: [{
+                        type: 'empty',
+                        prompt: 'Please enter your organisation'
+                    }]
+                },
                 username: {
                     identifier: 'username',
                     rules: [{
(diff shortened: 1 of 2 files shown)


dpaun:
For this to work, there should be some small changes in the user-service API. @TBoonX (or someone who has worked on the user-service) needs to make modifications to the '/register' handler, and add the 'organization' parameter.

abijames:
Update to user service is done - SWIK-676

kadevgraaf:
Hi @dpaun , @abijames , and @TBoonX 

Should we merge this in master - and thereby included it in the update to stable.slidewiki.org?

abijames:
@kadevgraaf  let me just remind myself what is on this branch...

dpaun:
I'm not sure. As far as I can see, it is already a part of stable.slidewiki.org, but it is not merged into platform.experimental.slidewiki.org.

abijames:
@dpaun is correct. These changes are already on stable.

bwulff:
This has been merged into branch SWAD-26_release_on_stable. Closing PR since we won't merge this into master as it is an interim solution as long as user groups right management is not in place.

We should keep it as an optional feature though! There may be adopters in the future who want to restrict signup in the way this does (e.g. only to their students)

PR #153, opened 2016-10-05
Swik-675 remaking user profile page

TBoonX:
Deleted tabs in my profile view
used thumbnail component in DeckCard

### DIFF
--- a/actions/user/userprofile/removeUser.js
+++ b/actions/user/userprofile/removeUser.js
@@ -18,8 +18,10 @@ export default function removeUser(context, payload, done) {
             } else
                 context.dispatch('DELETE_USER_FAILURE', err);
         } else {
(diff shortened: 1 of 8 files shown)


ali1k:
I still see a redundant '}' after the name of my deck! Could you remove that character?
Can you also check the 'Delete my account' functionality! if it is not implemented yet, can you hide it from the profile?

TBoonX:
I fixed these two bugs.
If a user visits the profile settings page an error is thrown - I will investigate this now

ali1k:
sounds good! let me know when it is ready!

TBoonX:
Its not an error. the console logs a 404 when a user have no decks, which is correct.
Ready to merge! - pulled master again

ali1k:
good! let's merge it then!

PR #154, opened 2016-10-05
Fix IE11 error - SCRIPT1014: Invalid character main.js 

huwf:
The IE issue was a third party identicons library (relied on by identicons-react) which we weren't doing anything on with babel, so it had template strings (with backticks) which IE didn't recognise.

This adds babel to the `node_modules` directory aside from a couple on which it breaks.  Modified in dev.config and prod.config.  Have also moved my webpack code from these files since it is no longer necessary from my point of view, and did the same from `packages.json`.


ali1k:
I was just trying to check this change, but now in my logs I see lots of warnings like:
`[BABEL] Note: The code generator has deoptimised the styling of "/Users/ali/my-nodes/slidewiki-platform/node_modules/immutable/dist/immutable.js" as it exceeds the max of "100KB".
[BABEL] Note: The code generator has deoptimised the styling of "/Users/ali/my-nodes/slidewiki-platform/node_modules/async/dist/async.js" as it exceeds the max of "100KB".
[BABEL] Note: The code generator has deoptimised the styling of "/Users/ali/my-nodes/slidewiki-platform/node_modules/lodash/lodash.js" as it exceeds the max of "100KB".`

Can you check the logs you get from `npm run dev` and `npm run build`?

huwf:
I got that as well, although it's just information rather than a warning.  According to http://stackoverflow.com/questions/35192796/babel-note-the-code-generator-has-deoptimised-the-styling-of-app-js-as-it-exc we can get rid of the message it if we  add a `compact:true` or `compact:false` to Babel depending on whether we want readable output.  I would suggest adding `true` to prod and `false` to dev?  Though it's only an issue in the files in `node_modules` I think

kadevgraaf:
Thanks for you investigation and fixes! 
Ali and me are still looking if there is a solution other than use of compact:true/false

PR #159, opened 2016-10-06
SWIK-692_delete_user_to_deactivate_user] 

TBoonX:
Remarks:

-     The other services are not aware of this! When using the frontend this is no problem.
-     Just login, deleteUser, getUser, getPublicUser handling this attribute. I think for the trial this is enough.
-     A user could be re-activated if someone access the database. There is no API for this


ali1k:
I think in the UI we should somehow show if the user is deactivated. Maybe use 'anonymous' in there, still needs to be discussed...

PR #160, opened 2016-10-06
SWIK-683 Download and PrintButton do the same:download the entire dec…

Sole-Valero:
As we don't have the pdf Service prepared to receive slides identifiers, we patch the bug by making that download button always gets a pdf with all the slides from a deck. Also, we do the same action for print button.
Now, both button call the pdfServce by:

`http://pdfservice.experimental.slidewiki.org/exportPDF/{deckId}`

It is possible to check it at: http://svalero.dsic.upv.es:3000/deck/8


kadevgraaf:
Some links still take the deck revision number, which breaks PDF service. Compare e.g., http://pdfservice.experimental.slidewiki.org/exportPDF/7-1 and http://pdfservice.experimental.slidewiki.org/exportPDF/7
 Can you fix this as well?

PR #178, opened 2016-10-17
SWIK-713_signin_on_signup

TBoonX:
Redirect in LoinModal after user is logged in and on signup or resetpassword page


TBoonX:
Abi tested it and closed the ticket.

PR #181, opened 2016-10-18
SWIK-731 now using PUT instead of POST against /resetPassword

bwulff:
Error description: When trying to reset a users password a message appears saying that the users mail address was not found in the database, even though the user account is existing.

This was caused because in services/user.js the call against the userservice's /passwordReset was implemented as a POST.

Fix: changing post to put.


rmeissn:
Is now using the right http method for reset password.

PR #182, opened 2016-10-20
Swik 732 fixing error actions refactoring

vinaymodi:
During code refactoring in Leipzig, some functions were not correctly refactored. Also fixed few dispatch strings. This was not making any side effect on the existing code because in ErrorStore, for all dispatch strings there is only one handler.


kadevgraaf:
@ali1k - Were you involved in this work during the Hackathon? If so, and if you have time, can you do a quick QA as well?

PR #187, opened 2016-10-25
Swik 642 validate comment title

dpaun:
This PR adds validation of comment (and reply) title and text. Currently, these fields are mandatory so if one of them is empty a notification will be displayed.


kadevgraaf:
Seems like a clear/good refactor of existing code.

PR #188, opened 2016-10-25
SWIK-457_language_dropdown_in_addeck

TBoonX:
Elaborated common/languagedropdown. used it in AddDeck


rmeissn:
Good additions and seems to work fine.

PR #192, opened 2016-10-27
SWIK-764 Sort my decks

rmeissn:
Added a sort button to the "My Decks" page.


TBoonX:
You missed handling null decks.

rmeissn:
That's impossible, because of https://github.com/slidewiki/slidewiki-platform/blob/master/actions/user/userprofile/fetchUserDecks.js#L15

PR #195, opened 2016-11-03
SWIK 724   Implement Slide Editor features of old SlideWiki.org – formulas, SVG editor, etc

kadevgraaf:
Does not work perfect in Safari, but is workable


kadevgraaf:
SVG editor needs more work.

PR #197, opened 2016-11-09
Swik 469 adding sources of slides

dpaun:
Functionality for saving and getting data sources from deck-service is implemented. This is closely related to the pull request [#43](https://github.com/slidewiki/deck-service/pull/43) in the deck-service repository.
-Data sources tab is enabled in the ContentModulesPanel.
-Add/edit/delete of data sources is enabled only for the owner of the selected slide.
-After adding/deleting a data source, the data sources count is updated.
-Deck data sources are collected from its slides.


kadevgraaf:
Hi Dejan,
Looks very good. Many options and nice functions!
When testing on my localhost I however noticed that the source are lost when you edit a slide and have a new revision. Do you have the same, or is this a configuration/localhost problem?
For aesthetics: the clickable area of the edit button on the right of each data source is a little bit shifted to the right in Google Chrome.
It seems Travis did not automatically build this PR into experimental.slidewiki.org for some reason, however, if you do another PR I expect this should work normally.

dpaun:
Hi, Klaas,
Thanks for the comments!
Actually, this, initial implementation connects data sources to slide revisions, so each revision has separate data sources. The idea for this came from [slide revision data model](https://slidewiki.atlassian.net/wiki/pages/viewpage.action?pageId=16285711) which includes the dataSources property. Should we change this and connect data sources to slides instead? Or make some changes to the code so that a new revision copies the previous revision's data sources?
I'll try to fix the edit button Chrome issue.

abijames:
I would suggest that sources should be copied over as each new version is made as they are linked to the slide content.

kadevgraaf:
Hi Dejan!

I think your initial implementation has the correct use of the slide revision data model. 
I agree with @abijames: changing/adding to the code so that a new revision copies the previous revision's data sources seems the correct option to me: then we also support revision of datasources. 
I.e., if someone edits a slide and removes/adds a data source (because the data source is (not) used in the slide after edit) - then in the new revision the data sources would be updated, and more importantly; in the previous revision you can still see the deleted data source that corresponds to slide content in that revision. 

Connecting datasources to slideRevisions (option 2 in your message) seems to have a lot of additional value over connecting data source to a slide (option 1). 
The data model already supports this.
I assume @mmeimaris, @ali1k, and @DaryaTarasowa whom worked on the revisions & data model, also agree. I'll tag them so they are up to date of this slideRevision usage.

Kind regards,

Klaas

dpaun:
Hi, Klaas!
I completely agree with you and Abi. I will implement this.
Best regards,
Dejan

mmeimaris:
Hi Abi, Klaas, Dejan. I completely agree with how the data sources should be treated when revising an item. @dpaun let me know if you want to collaborate on the implementation

PR #201, opened 2016-11-10
SWIK-739 social login

TBoonX:
I added  new components and actions.
I enhanced the UserProfileStore, UserRegistrationStore, UserRegistration, user, userprofile
Also a fix for cookies in in there.


kadevgraaf:
@rmeissn @TBoonX - Shall I or @bwulff QA this, for it to go in the update to stable?

TBoonX:
Klaas its yours. But you have to test it with the user-service branch.
See https://slidewiki.atlassian.net/browse/SWIK-161

kadevgraaf:
Hi Kurt!
Thanks. I talked with @bwulff  and we propose not to include the PR yet in the update to stable because it has a lot of new code/features on short notice/with little testing time. Would you agree or do you think it is quite safe and important to merge asap?

TBoonX:
Let's discuss this in Genua -  perhaps the trial partner really need this. But yes we have to test it plus it increases the deployment difficulty level. Thus just add it to stable if really necessary

abijames:
Also, as we are currently using the registration process to limit accounts to people associated with the project. we should only implement registration through social accounts when we have editing groups implemented.

rmeissn:
I've tested social login. At all: Good work! 
Here's what I've noticed:

* Is it really necessary to tell people that a new tab will open up? Other WebApps just open a new tab (or better: a new and small window).
* I was confused by the dialog that told me that login via social provider failed. --> Of course it failed, because I had never logged in via this provider before. But the text should sound more like "Hey, you have never tried to login via Facebook. We will now create a new account for you at SlideWiki. Is that what you intended to do or do you want to try logging in through another provider?"
* Please add popups to the dialog where I had to add extra information that state why i cannot use my e.g. email address in case it is already used by another account.
* If possible, please also fetch country, language, bio and organization from the social provider and/or add the language of my browser as language
* The "Authorized Accounts" section in settings confused me, because it seemed like I already enabled Google and Github, but Facebook is missing. Actually, it is vice versa. Enabled providers should be green and not enabled ones should be grey.
* I should not be able to disable Facebook login in case this is my only login option --> please disable the "disable" button in case this is the only provider left.
* I have no possibility to delete the account at SlideWiki. At least, I should be able to delete/deactivate my account.
* I was not able to login again through facebook after I've logged out on slidewiki. SlideWiki told me that I have never logged in through facebook and need to create an account.

TBoonX:
I thought it would be more accessible if there is a message. also it could take a few seconds until the opened tab closes itself - thus the user should be aware that he/she have to wait.
You are right, I have to remake the message for social login fail.
The popups of wrong fields are opened when the mouse is positioned over the input field. Do you want a error message box like in registration view?
biography is fetched but country is not possible beause I could get Germany, de, de_DE, Deutschland, .... I use for frontendLanguage always en_EN and for spokenLanguage the browser language.
OK I will change Integrations.js
Deactivate is now available. But we have to discuss how deactivated users have to be handled.
I couldn't validate your facebook failure. also with your credentials it just works.

rmeissn:
"It could take a few seconds until the opened tab closes itself - thus the user should be aware that he/she have to wait."
--> I think this is normal for most (if not all) providers, except that they typically open a new (small) window instead of a tab.
"The popups of wrong fields are opened when the mouse is positioned over the input field. Do you want a error message box like in registration view?"
--> Seems like I just missed it....will test it again in the next test cycle
"biography is fetched but country is not possible beause I could get Germany, de, de_DE, Deutschland, .... I use for frontendLanguage always en_EN and for spokenLanguage the browser language."
--> okay
"Deactivate is now available. But we have to discuss how deactivated users have to be handled."
--> feel free to ;)
"I couldn't validate your facebook failure. also with your credentials it just works."
--> Will test again in the next test cycle

actions/user/deleteSocialData.js
```diff
@@ -0,0 +1,4 @@
+export default function deleteSocialData(context, payload, done) {
```
rmeissn: You could add more than one method to an action (or file) by omitting the "default" and exporting several functions. Reimport is handled through `import {abc, xyz} from ...`

actions/user/registration/socialSignUp.js
```diff
@@ -0,0 +1,13 @@
+export default function socialSignUp(context, payload, done) {
+    context.service.create('user.socialregistration', payload, { timeout: 20 * 1000 }, (err, res) => {
+        if (err) {
+            context.dispatch('SOCIAL_CREATE_USER_FAILURE', err);
+            done();
```
rmeissn: According to [Fluxible Action Docs](https://github.com/yahoo/fluxible/blob/master/packages/fluxible/docs/api/Actions.md) you could omit the `done()` call in case you're using promises. Also have a look at this [Fluxible Actions Docs](http://fluxible.io/api/actions.html) for correct usage of promises with dispatch.
The following should work:
```
function socialSignUp(context, payload) {
....
  return context.dispatch(....);
....
}
```

components/Login/LoginModal.js
```diff
@@ -61,16 +67,61 @@ class LoginModal extends React.Component {
...
+        }
+    }
+
+    handleRegisterFirst(dismiss) {
+        if (dismiss === 'cancel')
+            return true;
+
+        localStorage.setItem(MODI, 'login_failed_register_now');
+
+        this.context.executeAction(deleteSocialData, { });
```
rmeissn: Wait for the first action to finish? E.g. use of promises?

components/Login/LoginModal.js
```diff
@@ -93,6 +144,124 @@ class LoginModal extends React.Component {
...
+
+            let win = window.open(url, '_blank', 'width='+width+',height='+height+',left='+left+',top='+topSpace+',toolbar=No,location=No,scrollbars=no,status=No,resizable=no,fullscreen=No');
+            win.focus();
+
+            return true;
+        })
+        .catch();
+    }
+
+    clickedFacebook(e) {
```
rmeissn: Use a parameter to distinguish between different providers instead of implementing three functions with the (nearly) exact same content.

components/Login/LoginModal.js
```diff
@@ -93,6 +144,124 @@ class LoginModal extends React.Component {
...
+            console.log('Error while parsing data', err);
+            return;
+        }
+        finally {
+            //delete data
+            localStorage.setItem(NAME, '');
+        }
+
+        //add language before send to service
+        let language = navigator.browserLanguage || navigator.language;
```
rmeissn: In case this is used elsewhere too, add it as a common function. Seems to me like a common function.

components/Login/LoginModal.js
```diff
@@ -93,6 +144,124 @@ class LoginModal extends React.Component {
...
+                confirmButtonText: 'Confirm',
+                confirmButtonClass: 'negative ui button',
+                buttonsStyling: false
+            }).then().catch();
+
+            return;
+        }
+
+        this.context.executeAction(newSocialData, data);
+        this.context.executeAction(userSocialSignIn, data);
```
rmeissn: Wait for the first action to finish? E.g. use promises.

components/Login/LoginModal.js
```diff
@@ -124,6 +293,12 @@ class LoginModal extends React.Component {
                         <div className="ui error message"/>
                       </form>
                       <br/>
+                      <div className="container">
+                        <i className="big circular facebook square link icon" onClick={this.clickedFacebook.bind(this)} ></i>
```
rmeissn: See parameter hint from above

components/Login/Social.js
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    findGetParameter(parameterName) {
+        let result = null,
+            index = 0;
+        location.search
+        .substr(1)
+            .split('&')
+            .forEach((item) => {
```
rmeissn: Use map+reduce if possible instead of a side effect (adding strings to result).

components/User/UserProfile/Integrations.js
```diff
@@ -0,0 +1,390 @@
...
+                buttonsStyling: false
+            }).then(() => {
+                this.context.executeAction(resetProviderStuff, {});
+                return true;
+            }).catch();
+        }
+        else if (this.props.UserProfileStore.addProviderError === false && nextProps.UserProfileStore.addProviderError) {
+            swal({
+                title: 'Error',
+                text: 'The provider couldn&apos;t be added. Unknown error.',
```
rmeissn: Giving a user an "unkonwn error" is not helpful. It just confuses the user.

components/User/UserProfile/Integrations.js
```diff
@@ -0,0 +1,390 @@
...
+
+        //prepare localStorage
+        localStorage.setItem(MODI, 'addProvider');
+        localStorage.setItem(NAME, '');
+
+        //observe storage
+        $(window).off('storage').on('storage', this.handleStorageEvent.bind(this));
+
+        //create new tab
+        let url = 'http://authorizationservice.manfredfris.ch:3000/connect/' + this.provider;
```
rmeissn: A hardcoded URL?!

components/User/UserProfile/Integrations.js
```diff
@@ -0,0 +1,390 @@
...
+            this.context.executeAction(updateProviderAction, '');
+            return false;
+        }
+        finally {
+            //delete data
+            localStorage.setItem(NAME, '');
+        }
+
+        //add language before send to service
+        let language = navigator.browserLanguage || navigator.language;
```
rmeissn: The whole paragraph seems to be repeated in another file --> export into a function --> avoids copy and paste

components/User/UserProfile/Integrations.js
```diff
@@ -0,0 +1,390 @@
...
+        if ( (data.token.length < 1)
+          || (data.provider.length < 3)
+          || (data.token_creation.length < 22) )
+            //Failure
+        {
+            this.context.executeAction(updateProviderAction, '');
+            return false;
+        }
+
+        if  (data.email.indexOf('@') === -1 || data.email.indexOf('.') === -1 || data.email.length < 5) {
```
rmeissn: A regex (search on stackoverflow) might be better to check for a valid email adress

components/User/UserProfile/Integrations.js
```diff
@@ -0,0 +1,390 @@
...
+                        google = true;
+                        break;
+                    case 'github':
+                        github = true;
+                        break;
+                }
+            });
+        // console.log('Integrations render()', this.props.UserProfileStore.user.providers);
+
+        let facebook_icon_classes = classNames({
```
rmeissn: all of the following statements are either true or false. You might be able to condense rules to fewer lines by using `!facebook` or `facebook` more

components/User/UserProfile/Integrations.js
```diff
@@ -0,0 +1,390 @@
...
+                    </p>
+                  </div>
+
+              </div>
+              <div className="ui segments">
+
+                  <div className="ui secondary segment">
+                      <h3>Login Provider</h3>
+                  </div>
+                  <div className="ui segment">
```
rmeissn: The following paragraphs have the exact same structure. Use a loop to cycle through the list of all providers (currently three) instead writing it three times.

components/User/UserProfile/UserProfile.js
```diff
@@ -117,24 +120,31 @@ class UserProfile extends React.Component {
     }
 
     displayAccounts() {
-        return (<div>
-            <div className="ui segments">
-            <div className="ui secondary segment">
-              <h3>Change password</h3>
-            </div>
+        let changePassword = '';
```
rmeissn: Use an inline if else statement .... `let tmp = (condition) ? 'yes' : 'no' `

components/User/UserProfile/UserProfile.js
```diff
@@ -143,8 +153,14 @@ class UserProfile extends React.Component {
         return (<PrivatePublicUserProfile user={this.props.UserProfileStore.user} decks={this.props.UserProfileStore.userDecks}/>);
     }
 
+    displayIntegrations() {
```
rmeissn: If it's that easy, you might be able to omit the function at all by moving `<Integrations />` upwards.

components/User/UserRegistration/UserRegistration.js
```diff
@@ -116,10 +134,55 @@ class UserRegistration extends React.Component {
             return (usernameNotAllowed !== undefined) ? !usernameNotAllowed : true;
         });
 
-        $('.ui.form').form(validationRules);
+        $('.ui.form').not('.registrationmodalform').form(validationRules);
```
rmeissn: Use react identifiers instead of searching for classes! Searching for classes caused problems before.

components/User/UserRegistration/UserRegistration.js
```diff
@@ -116,10 +134,55 @@ class UserRegistration extends React.Component {
...
+            })
+            .then((dismiss) => {
+                if (dismiss === 'cancel')
+                    return true;
+
+                this.context.executeAction(navigateAction, {
+                    url: '/'
+                });
+
+                $('.ui.login.modal').modal('show');
```
rmeissn: Use react identifiers.

components/User/UserRegistration/UserRegistration.js
```diff
@@ -116,10 +134,55 @@ class UserRegistration extends React.Component {
...
+
+                return true;
+            })
+            .catch(() => {
+                $('.ui.socialregistration.modal').modal('hide');
```
rmeissn: Use react identifiers.

components/User/UserRegistration/UserRegistration.js
```diff
@@ -170,21 +233,32 @@ class UserRegistration extends React.Component {
...
+                firstname: this.refs.firstname.value,
+                lastname: this.refs.lastname.value,
+                username: this.refs.username.value,
+                language: language,
+                email: this.refs.email.value,
+                password: hashPassword(this.refs.password.value),
+                grecaptcharesponse: this.state.grecaptcharesponse
+            };
+        } catch (e) {
+            //somehow this is sometimes called when doing login
```
rmeissn: Please debug it instead of having a try catch statement

components/User/UserRegistration/UserRegistration.js
```diff
@@ -208,6 +282,160 @@ class UserRegistration extends React.Component {
...
+        localStorage.setItem(NAME, '');
+
+        this.provider = provider;
+
+        //observe storage
+        $(window).off('storage').on('storage', this.handleStorageEvent.bind(this));
+
+        //show hint before open tab
+        const Provider = this.getProviderName();
+        swal({
```
rmeissn: Maybe not needed at all? Just open the login window.

components/User/UserRegistration/UserRegistration.js
```diff
@@ -208,6 +282,160 @@ class UserRegistration extends React.Component {
...
+
+            let win = window.open(url, '_blank', 'width='+width+',height='+height+',left='+left+',top='+topSpace+',toolbar=No,location=No,scrollbars=no,status=No,resizable=no,fullscreen=No');
+            win.focus();
+
+            return true;
+        })
+        .catch();
+    }
+
+    clickedFacebook(e) {
```
rmeissn: Copied and pasted from above ... export to an external function/file

components/User/UserRegistration/UserRegistration.js
```diff
@@ -296,9 +525,23 @@ class UserRegistration extends React.Component {
                         <div >
                             By clicking Sign Up, you agree to our <a href="">Terms</a>.
                         </div>
+                        <div className="ui dividing header" ></div>
+                        <div className="container">
+                            Register with another platform:
+                            <br/>
+                            <i className="big circular facebook square link icon" onClick={this.clickedFacebook.bind(this)} ></i>
```
rmeissn: User parameters (like already written above)

components/User/UserRegistration/UserRegistrationSocial.js
```diff
@@ -0,0 +1,254 @@
...
+import ReactDOM from 'react-dom';
+import classNames from 'classnames';
+import {navigateAction} from 'fluxible-router';
+import {connectToStores} from 'fluxible-addons-react';
+import socialSignUp from '../../../actions/user/registration/socialSignUp';
+import checkEmail from '../../../actions/user/registration/checkEmail';
+import checkUsername from '../../../actions/user/registration/checkUsername';
+import UserRegistrationStore from '../../../stores/UserRegistrationStore';
+
+const headerStyle = {
```
rmeissn: I saw this before ... copy and pasted. Avoid it by exporting it.

components/User/UserRegistration/UserRegistrationSocial.js
```diff
@@ -0,0 +1,254 @@
...
+        e.preventDefault();
+
+        $('.ui.socialregistration.modal').modal('hide');
+        let user = this.props.UserRegistrationStore.socialuserdata;
+        user.email = this.refs.email.value;
+        user.username = this.refs.username.value;
+        user.forename = this.refs.firstname.value;
+        user.surname = this.refs.lastname.value;
+
+        let language = navigator.browserLanguage ? navigator.browserLanguage : navigator.language;
```
rmeissn: Copied and pasted from above ... export to an external function/file

services/user.js
```diff
@@ -44,6 +44,31 @@ export default {
                         error: err
                     });
                 });
+        } else if (resource === 'user.socialsignin') {
+            rp.post({
+                uri: Microservices.user.uri + '/social/login',
+                body: JSON.stringify({
```
rmeissn: Just stringify args instead?!

stores/UserProfileStore.js
```diff
@@ -32,6 +32,11 @@ class UserProfileStore extends BaseStore {
         this.jwt = '';
         this.userpicture = undefined;
         this.errorMessage = '';
+        this.socialLoginError = false;
```
rmeissn: use an object instead of several independent items

stores/UserProfileStore.js
```diff
@@ -198,6 +218,11 @@ class UserProfileStore extends BaseStore {
         this.emitChange();
     }
 
+    handleSocialSignInError(err) {
+        this.socialLoginError = true;
```
rmeissn: You might be able to reset it immediately after `this.emitChange()` instead of having a full component --> action --> store cycle again.

stores/UserProfileStore.js
```diff
@@ -207,6 +232,68 @@ class UserProfileStore extends BaseStore {
...
         return message2;
     }
+
+    deleteSocialData() {
+        this.socialLoginError = false;
+        this.emitChange();
+    }
+
+    socialRegister(res) {
+        this.userpicture = res.picture;
```
rmeissn: Why is this needed?

components/User/UserProfile/Integrations.js
TBoonX: As I have three flags its better for readers to not change it

components/User/UserProfile/UserProfile.js
TBoonX: I will not do this, because this will break the used style

services/user.js
TBoonX: I tried it but the service acted very strange. As the other calls to this service also use stringify I will not change it. Also the parameter "json: true" just do the stringify for you, thus no overhead.

stores/UserProfileStore.js
TBoonX: Done.

components/User/UserRegistration/UserRegistration.js
rmeissn: Hm, good question. It should be.....try to search in the React documentation.

components/User/UserProfile/Integrations.js
rmeissn: Depends on your attitude ;) For me, it's easier to read in parameters and I'm confused by repeated functions that do the same thing :P

rmeissn:
Executed a functional test together with Kurt. We found some bugs that need to be fixed before it may be merged.

rmeissn:
Executed server side tests, everything's fine, ready to merge

PR #203, opened 2016-11-10
Swik 686 autocomplete

schatzopoulos:


kadevgraaf:
Looks good!
Can you also easily implement this in the search-box in the top/header menu (which is visible everywhere)?

PR #206, opened 2016-11-11
SWIK-762

Sole-Valero:
Added a sweet alert to receive the confirmation from user before deleting an slide. 
Code added to deleteTreeNodeWithRevisionCheck

Related with https://slidewiki.atlassian.net/browse/SWIK-762

After the user press the delete button, it should appear an alert message to get the final confirmation from user. In case the change requieres the creation of a new revision of the deck,  the user will ask previously showing only a confirmation about it will create a new revision of the deck...nothing about the deleting process. I take this decision in order not to chain two alert messages


abijames:
Happy for this to merge

PR #210, opened 2016-11-14
Add Submit button to tab order. Added accessible labels to form fields

abijames:
Changes to the search results page to solve some of the keyboard and accessibility issues.


kadevgraaf:
Hi Abi! 
This broke the autocomplete in the top-menu search field on my localhost, but after updating from master it works 👍 
Cheers,
Klaas

PR #212, opened 2016-11-15
restore label for header search box

schatzopoulos:


abijames:
This branch now has the correct labels so happy for it to be merged.

PR #225, opened 2016-12-01
SWIK-804-header-responsive-design

rmeissn:
Added a mobile better menu/header on mobile devices via media queries.

Needs still some testing and code is ugly, but I have no idea how to improve the ugly code.


abijames:
@rmeissn I have tried testing this but can't see any difference in the header on different browser sizes. I am investigating if I can access my localhost on my phone to do a real test. But in the meantime @kadevgraaf do you want to have a look at this too? If we think the code is ok may be easier to merge to experimental test further on that.

kadevgraaf:
Code is OK and works on my local machine.

rmeissn:
@abijames I guess it will not work when you only resize your browser, because it's a reactish implementation that updates only when the component changes it's state. It works for me, in case I use Chromes "emulate mobile device" feature.

PR #226, opened 2016-12-01
SWIK-743 notifications usergroups

TBoonX:
Based on SWIK-644 which is not yet a PR.
Added two new types of notifications.
Improved reading userid.


dpaun:
The code looks ok.

PR #227, opened 2016-12-02
Swik 790 render remaining components when service error

vinaymodi:
503 service errors won't appear as a separate page. This can be merged into stable. But it needs some more work to show 503 service errors (upstream server errors) at some other place on the deck page component therefore this branch should not be deleted.


kadevgraaf:
Looks good. Nice ServiceErrorStore 👍 
I do get _Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) v data-reactid="2"><div class="ui invert
 (server) v data-reactid="2"><!-- react-empty: 3 -_

But this might be because I am running it on localhost.

kadevgraaf:
I am not sure if this is best for users - now there is not any error (the screen freezes) - unless we fix the errors, or shown an error message nevertheless. @abijames what do you think?

PR #232, opened 2016-12-02
tags search field disabled

schatzopoulos:
tags field in search is disabled as no tags are supported yet.


abijames:
@schatzopoulos can you remove the disabled Tags field from the tab index (tabIndex=-1). Also, the user field has a different label (for="search_id") from the input field (id="users_input_div") which is causing an accessibility error.

abijames:
@kadevgraaf my local version is not loading :-( can you test this? just need to make sure you can't tab to the tags field in the search UI....

abijames:
tested and tab order now correct so this can be merged.

PR #236, opened 2016-12-08
Fix for slides larger and smaller than 960*720

kadevgraaf:
test


kadevgraaf:
Also needs to be implemented for thumbnail and print service

PR #240, opened 2016-12-09
Swik 754 prototype drag and drop slide reordering

stavmars:
- Changed the navigate logic after reordering based on @abijames suggestions

- Added "Refreshing deck structure" sweet alert message while reordering as a temporary way to prevent the user from initiating another deck tree operation before the deck service is updated.


abijames:
works well! The alert is nicely timed

PR #242, opened 2016-12-12
Swik 802 investigate implement import of diagrams from pptx to slide wiki

sednanref:
Added support for many types of charts on slideViewPanel (working perfectly), slideEditPanel and slidePresentation (with some bugs yet, specially with the resizing).


kadevgraaf:
Looks good!
In several browsers (Chrome, FF, Safari), or perhaps due to a slow connection, the charts are not fully loaded in the slide edit panel - if you then save the slide, the chart will also appear partially in the slide view panel (because parts that were not loaded are not saved in edit mode).

kadevgraaf:
Hi @sednanref ! 

W.r.t. including d3 library: 
+                <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.17/d3.min.js"></script> {/* D3 */}
+                <script src="https://cdnjs.cloudflare.com/ajax/libs/nvd3/1.8.1/nv.d3.min.js"></script> {/* NVD3 */} 

For performance and to remove outside dependency to cdnjs, do you want to include the nv.d3 and d3 libraries from the slidewiki-platform repository itself, e.g., from custom_modules or better: using npm / package.json 

Another performance improvement might be to detect if there are any diagram elements in slide content before doing the forceUpdate and calling the  ChartRender.createCharts(); on various pages.

sednanref:
@kadevgraaf added npm modules :) it was a bit tricky because of some uncompatibilities between de npm modules of d3 and nvd3 in some versions, also did the check before rendering the chart. The bug in the slide Edition mode is still there... I'm trying to figure out how to resolve it :)

PR #243, opened 2016-12-14
Swik 516 bug editing mode and moving different slide should be warned

sednanref:
Adding the onbeforeUnload warning in the slideEdit mode when dragging or resizing a content box.


kadevgraaf:
Good solution. Works well.

PR #244, opened 2016-12-15
SWIK-887_last_slide_deletion_could_fail

TBoonX:
just corrected checking of count in tree after delete


TBoonX:
This time changes? in semantic-ui were added. strange - should I revert these?

kadevgraaf:
Works well on my localhost.

PR #246, opened 2016-12-16
Swik 882 update search results after autocomplete selection

schatzopoulos:


kadevgraaf:
Works good and is fast.

PR #248, opened 2016-12-19
Improved accessibility on user settings page.

abijames:
Added for/id to form fields in user settings page. Language/country dropdown still not correctly labelled & will need to be fixed with
Semantic-UI-React. Removed Uplad Image button from tab order.


kadevgraaf:
Not sure how to test for/id - but everything seems to be working and keyboard accessible.

PR #251, opened 2017-01-10
SWIK-908 Remove AddDeck button double hover visuals

txwkx:
I got rid of the **AddDeckPanel** component, since it was returning 1 line of plain HMTL and imported in only 1 other component. In case it should be used somewhere else in the future it could be easily subtracted into a stateless functional component.

_Side effects_: The height of the header slightly increased due to css rules of `class="item"`. 
I strictly focused on the fix by only manipulating the markup.


abijames:
Hi @txwkx when you tab through the header, you have to tab twice on the add deck button. A screenreader reads the button correctly on the first tab, there is silence on the 2nd so this is the one that should be not foccussable . This was an existing issue I hadn't identified a solution for but is still present on your branch. Please IM me on slack if you need more detail.

abijames:
Checked latest version. All works well so happy for this to be merged

bwulff:
@abijames @TBoonX thanks for looking at this. And of course cheers to @txwkx to your first PR!

PR #252, opened 2017-01-10
Redirecting the user to the view mode of a deck/slide if he was in ed…

sednanref:
…ition mode and signed  out.


kadevgraaf:
Works well - user goes to correct URL. Thanks!

PR #256, opened 2017-01-16
Fix history tab action not working when navigating directly to nested…

kprist:
This fixes the issue for the history tab alone. The cause is the missing spath, and in JIRA it is implicit that it may cause bugs in other components, but haven't been able to locate any. 

This fix only works under the assumption that the deck tree is already loaded in the page as well, which under the current design is always true.


kadevgraaf:
👍

PR #257, opened 2017-01-17
Swik 839 sort search results

schatzopoulos:
Brings functionality for sorting search results based on relevance and last updated date.


kadevgraaf:
Hi @schatzopoulos !
Looks good! 
I noticed a small thing: after you select "last updated" as sort param, the results are correctly sorted, but the button for selecting search params still defaults to "relevance" (despite the check at https://github.com/slidewiki/slidewiki-platform/pull/257/files#diff-b7bd206df2f1a3f2234693dfe82c13c9R27).
Other than that I am ok with merging this 👍 
@abijames - is this ok w.r.t. usability/accessibility?

abijames:
Hi @schatzopoulos - the sort button needs some more markup to improve accessibility:
role="button"
aria-haspopup="true"
aria-label="Sort by"
... this doesn't fix everything but is enough until we move to semantic-ui-react

For usability, the exchange icon doesn't work as left/right arrows don't indicate the order is going to change. I'd suggest either "List Layout" or "Sort Content Ascending".

I know this replicated the My Decks page so  this applies to the My Decks page as well  @rmeissn

schatzopoulos:
@kadevgraaf Are you sure that it defaults to "relevance"? I 've tested it in Ubuntu with Chrome/Firefox and seems to be working ok... Can you check again? 
@abijames Can you check if the accessibility issues mentioned are resolved?
Thanks

kadevgraaf:
@schatzopoulos - works for me now :) Maybe I did not refresh properly
I'm ok for this to be merge once Abi is happy with accessibility.

abijames:
Ok to merge

PR #260, opened 2017-01-20
Swik 921 upgrade ckeditor to 42

kprist:
This makes somewhat significant changes to the way ckeditor lib is set up: 

1) the configuration file used is not the default, it is in assets and can be modified independently
2) ckeditor lib is added in package.json and can be upgraded with npm
3) external plugins are also handled differently:
  a) youtube plugin has npm package, and has been also upgraded from 2.0.9 to 2.1.0
  b) lineheight plugin is fetched from its git repo (master branch). npm integration via napa dev dependency: https://github.com/shama/napa
  c) texzilla plugin files were copied from old location in custom_modules/ckeditor/plugins/texzilla to custom_modules/texzilla 

Also, the upgrade to 4.2.6 includes some very obvious aesthetic changes.


kprist:
Just realized the texzilla plugin does not work after the ckeditor upgrade, I will need to rebuild it from its source.

kadevgraaf:
Works well as far as I can see!
Copying from MS word works nicely now. 
Copying directly from Powerpoint gives a file-upload error, but we might be able to fix this (as this is a new feature we do not introduce a new bug here I think).
Some image editing bugs still remain, but I cannot see additional ones being introduced.
Nested bullet and numbered lists also seem to work - see e.g. http://localhost:3000/deck/7-3/slide/111-4/111-4:8
Ckeditor also remembers style, font-size and lineheight after editing text.
Youtube plugin works. code highlighing. 
Texzilla works like before (without same formula in dialog though), only after editing an existing formula it gets placed outside the input boxes - perhaps we replace with MathJax in time.
Install requires "npm install napa --save-dev" as well - not sure how this affects CI/CD process - @bwulff  @TBoonX @rmeissn .
I am OK with this being merged for us to continue working from this version.
@abijames - do you think we can go ahead? Is it still accessible? I will change the color to default (as shown on CKeditor) because now the blue color does not look good. Also I will work on having a toolbar instead of the inline editor.

abijames:
@kadevgraaf @kprist - I have just tried adding new slides and new deck on experimental and CKeditor is not loading.

kprist:
I have reported this on JIRA as well: 


> It seems this doesn't work properly on experimental: ckeditor won't load and the console logs:
> Error: CKEDITOR.resourceManager.load Resource name "lineheight" was not found at "/ckeditor-plugins/lineheight/plugin.js?t=H0CF"
> This is probalbly related to a concern Klaas Andries de Graaf expressed at some point, whether the use of the 'napa' dev dependency in package.json would interfere with build and deploying.
> About the napa package: it is a dev only package that is used during 'npm install' phase that helps use a regular (non-npm packaged) git repo as an npm package. I used that in order to use the lineheight plugin directly from the source git repo ( https://github.com/lotusgodkk/lineheight ).
> I assume the reason it is not working is some failure during 'npm install' ( which I assume is part of regular deployment process ), and for some reason the relevant git repo is not locally cloned. Or the napa package is not properly installed?
> I tested the current package.json locally by removing the node_modules/ directory and running npm install and it works fine. I am not sure how else to test the build/deploy process.

kadevgraaf:
Shall we merge https://github.com/slidewiki/slidewiki-platform/pull/266 ? It contains the fix.
We can later re-merge the branch if more fixes are added.

PR #265, opened 2017-01-23
SWIK-879_404_public_profile

TBoonX:
Added check in public profile if user is equals signed in user


abijames:
Tested. This is ok to merge.

PR #266, opened 2017-01-23
Replace use of texzilla plugin with MathJax in CKEditor

kprist:


kadevgraaf:
Hi @kprist 
Thanks, looks good!
Mathjax rendering is not yet included on https://github.com/slidewiki/slidewiki-platform/blob/master/components/Deck/ContentPanel/SlideModes/SlideEditPanel/SlideContentEditor.js - users would see the source of the formulas, but not how they would visually show on the screen.
I tried to add this on my localhost, but I cannot get it to work quickly - do you know a good solution for this?

PR #268, opened 2017-01-24
Swik 930 slide revision switch by anonymous user

stavmars:
This PR hides the revert button for not logged in users.


kadevgraaf:
Works well on localhost 👍

abijames:
@stavmars functionally this works. But there is nothing in the UI to show the user which revision they are viewing. Can you implement something like the unhide icon next to the revision that is currently being displayed.

Also, when signed in (but not the owner) I can change the active revision.

abijames:
@stavmars @kadevgraaf shall we merge this anyway and make new tasks for the additional functionality?

kadevgraaf:
Sounds good. This is already a nice improvement.

stavmars:
@abijames  yes, I think that would be best.

PR #271, opened 2017-01-27
SWIK-633 Fix render issue of language specific characters

txwkx:
The DB stores the data in the following manner:
`&#xBF;` for `¿`

Simple data type casting on the client did the trick.

> A test deck could be found in the comment section of the respective ticket


kadevgraaf:
Works well! Thanks!

PR #285, opened 2017-02-17
Swik 961 fixing wrong activity feed and counters after back button

dpaun:
This PR solves a problem with Chrome (and possibly some other browsers) caching the initial version of data (e.g. comments count) (before changes) which is then used when the Back browser button is pressed (SWIK-961).
It uses 'localStorage' to detect cases when invalid data have been read from the browser cache. When this is detected, additional call(s) to micro-service(s) are/is executed.
Invalid data still may be briefly displayed, but I don't think this can be avoided without disabling caching.


txwkx:
Checked it using Chrome -v 56
Works :shipit:

kadevgraaf:
Solves the bug for me as well in Chrome 56.0.2924.87 (64-bit)

PR #291, opened 2017-02-17
Swik-986_UI_INTL

DaryaTarasowa:
Code refactoring, INTL architecture


kadevgraaf:
Hi Darya,
How can I test this with a locale different from English?
I see there is support for russian locale and I see that there is setlocal action and component, but I cannot find an URL to navigate to this.
In my settings I cannot change the Interface language to anything other than English.

ali1k:
We should make sure that all the changes are compliant with our current webpack build mode. I would postpone merging this until we have added more locales and have done more tests on how it exactly works. I already talked to Darya about that. We need more documentation on ui intl for developers how it works and how to use it in our components.

kadevgraaf:
W.r.t. planning -> the original/current slidewiki.org does not have localization, so we should continue work on this after migrating to slidewiki.org around June .

abijames:
@kadevgraaf @ali trials require this function to live before September. I anticipate we will then need to localise guides etc so it would be good to have this ready sometime before that deadline. Please can we discuss this further at the meeting next week as I would like to understand how it will work with text in the code that can be verbalised to users (e.g. alt-tags, hidden labels). I will attempt to do some reading up on possible strategies.

PR #303, opened 2017-02-27
    SWIK-799      Add zoom out/zoom in button/function in reveal.js Add zoom-js plugin to initizalize

kadevgraaf:


kadevgraaf:
Press Alt-click to zoom in

PR #307, opened 2017-03-02
SWIK-1033-deadlock-user-service-unavailable

rmeissn:
I've implemented a quick fix that temporarily solves the problem (still old architecture/implementation). This may be merged in order to have a quick-fix of the problem. Please do not delete the branch.

I want to refactor (find a better solution) the whole functionality with Ali Khalili to get rid of the problem in the first case and to have it more resilient against issues.


TBoonX:
For me the **lifelock** is gone.
But when I am logged in, viewing the user profile settings, the service gets unavailable and then reload the page, then there is no redirect. Is this a bug of the profile component or is a redirect missing?

rmeissn:
@TBoonX I am able to reproduce your comment. This is indeed a bug, but another one (and possibly a complex one). Can you please open a new bug ticket?

PR #314, opened 2017-03-08
SWIK 814 tag tab

kadevgraaf:


kadevgraaf:
Hi @aw3s0me 
The tag-tab looks nice!
I merged master into this and resolved some conflicts.
When I now run it locally and try to save a slide it however gives the following error:
body: '{"statusCode":400,"error":"Bad Request","message":"child \\"tags\\" fails because [\\"tags\\" must be an array]","validation":{"source":"payload","keys":["tags"]}}' } }
Can you fix this so we can then merge it to platform.experimental.slidewiki.org ?

PR #319, opened 2017-03-14
SWIK-1061 Exchanged the thumbnail retrievel code to use the new file-service API

rmeissn:
I've removed the image-service from the docker-compose and microservices.config file, as the service will be discarded (stopped & removed). I've changed the thumbnail retrievel code to match the new API of the file-service.

I haven't tested the changes yet, as I wait for approval of slidewiki/file-service#2 and https://github.com/slidewiki/deck-service/pull/63 first. I'll comment this PR again as soon as the other PR has been merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,14 +33,3 @@ add the config for your microservices:
 `npm run dev`
 
 check server at `localhost:3000`
-
-
-## Migrate thumbnails ##
-```
-cd /data/files
-find ./ -name thumbnails -type d -exec find {} -name *.png \; > list
-while read in; do cp "$in" /data/files/slideThumbnails/ ; done < list
-cd slideThumbnails
-for i in *.png ; do convert "$i" "${i%.*}.jpeg" ; done
-find . -name "*.png" -type f -delete
-```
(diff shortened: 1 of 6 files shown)


rmeissn:
Recent versions of deck-service and file-service are online and working. Please test that correct thumbnails are shown.
## How to test
- Go to http://localhost:3000/deck/1066 and test if thumbnails are visible
- Upload a new deck, wait a few secs and check if thumbnails are present (both, on the "my decks" page and the deck overview)

kadevgraaf:
I cannot view the mydeck page, so that is hard to test. This is also the case on experimental, so probably not caused by this code.
Seems to work for upload (http://localhost:3000/deck/1067) and viewing other decks, except for https://platform.experimental.slidewiki.org/deck/1066
I am fine with this merge if it is backwards compatible with old thumbnails.

rmeissn:
I am not able to replicate your "My Decks" page issue on my local deployment and also on experimental. This is probably a different error you need to dig into.

Changes are not really backward compatible, but I migrated all existing thumbnails on experimental to the new infrastructure. We would need to execute the instructions (added to the Readme of the file-service) once on testing and once on stable (as soon as a new version is deployed).

PR #320, opened 2017-03-14
842 implement reporting function/button/dialog

sednanref:
there is now a functional modal to send reports. The reports will be handled (by now) using the client's mail.


kadevgraaf:
Looks good / clear UI.
If I test it in Chrome and Thunderbird, the thunderbird "TO" field will be filled with:
_ka.de.graaf@vu.nl&subject= [SlideWiki] Report on Deck/Slide&body=Report made by user2 Report on Slide10319-1 From Deck1068-1 Reason of the reportspam Description of the reportKlaas is spamming with test decks. This is allowed_ = it is not distributed across the subject and body field in Thunderbird.

sednanref:
Hi Klaas! I do use Thunderbird and Chrome too, but the function gives me the following: 
screenshot from 2017-03-17 14-09-47 could it be an issue regarding thunderbird or chrome versions?

kadevgraaf:
Perhaps indeed. I tried with a different message in a different browser (firefox+ thunderbird on OS X) but the same happens. I also tried with a new Thunderbird version, but to no avail :)

sednanref:
Hi Klaas, I modified the string, it seems that I used a symbol incorrectly in it  (I still don't know why it worked on my machine), but I'm not sure if it is working for everyone, can you try again :)? thanks!

kadevgraaf:
Hi @sednanref It works for me now, thanks!

sednanref:
Great!

PR #321, opened 2017-03-15
Update fork button for new forking

stavmars:
This PR makes the necessary changes in order for the fork button to work with the updated deck-service fork.


kadevgraaf:
Works well!
For future dev perhaps we should have a confirmation for the fork button (as in original slidewiki.org I recall).

PR #322, opened 2017-03-16
Cookie banner added

urashidmir:
Cookie banner added


kadevgraaf:
Works well.
Note: In webpack/dev-server.js and dev.config.js 'use strict' was added to properly run webpack on Umar's local machine - from what I read about 'use strict' I expect this also to work on experimental and other machines.

PR #323, opened 2017-03-20
Swik 378 import odp

dpaun:
This PR introduces just a few changes to enable upload of ODP files.
A couple of labels have been modified to show that it's possible to import ODP files.
It also fixes the problem with default deckId (165) , which is used in case of some errors on the import-service.
This is related to the 'SWIK-378-Import_ODP' branch on the import-service which has already been merged.


dpaun:
I've checked the SendReportStore. It really doesn't have the creationSuccess method. Perhaps @sednanref simply forgot to delete these two handlers (CREATION_FAILURE and CREATION_SUCCESS).

sednanref:
Hi @dpaun you are right, I'll fix this and send a pull request from that branch! My bad

sednanref:
PR sent

kadevgraaf:
It works now (y)
Thanks @dpaun for tracing down the method and thanks @sednanref for fixing it!

PR #342, opened 2017-03-29
SWIK-1005-user-image-upload

rmeissn:
I've added the functionality and it's legen ..... wait for it ..... dary! Because it's all client-side =)


kadevgraaf:
Perfect! Looks very good and works nicely!

TBoonX:
Please add fileendings checks before loading it into the crop tool.
Also you should check if crop was successfully - I was able to use a zip file ^^

rmeissn:
Awwww, Kurt....you should know that a zip file is not a picture? Or do you wanted a cool matrix style profile picture by using random data? :D
--> fixed

abijames:
There are quite a few accessibility issues. I have added notes to the JIRA ticket. If we can address the keyboard accessibility of the buttons and modal window then we can address cropping via the keyboard in the future

rmeissn:
@abijames I've exchanged the semantic-ui modal with a react-semantic-ui modal. Can you please retest if accessibility (of the modal) is better now?

@Sole-Valero Did I implemented the modal right? I'm getting an error in the browser console that server and client side rendering are resulting in different DOM.

> warning.js:36 Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) ></div></div></div><button id="ChangePic
 (server) ></div></div></div><!-- react-empty: 135

@Sole-Valero Is there an option to move the modal action buttons to the right side of the modal, instead of the left side?

Sole-Valero:
In my case, the buttons appear on the center...because I put them inside the first segment for the content field. Try to add the style prop with the tex aling to the left

rmeissn:
Updated the modal according to Soles recommendations. In order to be in line with SweetAlert2 modals, I moved the heading to the left and the buttons to the right (as the default semantic-ui modal would look like too). Is this a good decision, @abijames or what is our desired style?

abijames:
Style of modal is ok. Can the focus default to the save button so that you can press enter to save?

I found that in the photo selector dialog, if I pressed Enter after selecting a photo then the dialog was cancelled. As the Open button was highlighted this was not the expected behaviour.

I am still concerned that there is no way to upload an image without drag and dropping the crop box. Its also nearly impossible to do on a touch screen. One option would be to show the modal window and have a button to select to crop the photo and then load the cropping tool. This would allow users to choose to upload without cropping.

Finally if you load up a large photo then the modal is larger than the screen. Here is a screenshot from a 12inch display. You can see that half the modal is cut off and there is a long way to scroll to get to the buttons. 
image

rmeissn:
@abijames 
I changed the initial focus to the save button and tried to improve the image-size. It turned out to be much more complex than I thought, so I implemented something that acts as a quick workaround.

About the file-dialog: I tried to replicate the behavior you described, but I can't. Selecting an image on both, Chrome and Firefox, and hitting enter opens up the modal to crop the image. As this dialog is provided and handled by your browser/system, it can't be manipulated. The default selection for "cancel" might be a default for your system or a bug of your browser.

As the library seems to introduce some more issues (frame size, cropper size, touch support, ...), I would need to either improve it myself or use another one (like https://www.npmjs.com/package/react-image-crop ). Improving it might be less effort than using another library, as the current one provides many features I would need to reinvent/reimplement for the other libraries.

Maybe we can get to a trade-off: I'll create an issue at the Github project of the library and ask for some features. Hopefully the main developer has time and implements these features or points me to some workarounds. We do merge the feature for now with the current library and introduce a ticket in Jira to improve the feature in the future (containing a link to Github issue I created and to alternative libraries). Would that be okay?

abijames:
@rmeissn if you removed the MaxWidth of the photo (in this line
         payload.picture = this.refs.cropper.crop({ maxWidth: 170 });
Or set the maxwidth to 100% else where would that set the default crop size to the whole image?

rmeissn:
@abijames I already tried to change some CSS and to use some of the library properties (e.g. `width="100%"`, but neither if it worked. In contrast, this introduced strange errors (like crop box outside of the picture) that are way worse than having to resize the crop box. I am writing an issue for the lib anyway, so I'll ask the developer about it.
The line you were mentioning (`payload.picture = this.refs.cropper.crop({ maxWidth: 170 });`) sets the width of the cropped picture, that will be saved (we'll save as 170px x 170px). This won't change the crop box width/height.

rmeissn:
Abi confirmed to merge it.

Sole-Valero:
I couldn't test it because experimental doesn't works...but if Abi says ok...it should be perfect

rmeissn:
@Sole-Valero The modal seems to work like a charm. There are "just" some accessibility issues about the cropper I used (third party library). But these must be fixed at the library itself (I already posted issues to the lib and might implement some missing features myself).

PR #344, opened 2017-04-03
SWIK-751_slow_internet_signin_form

TBoonX:
Removed clearing input (will be empty on success because reload is triggered) and added making inputs disabled and loading as long as request needs.
Used component state for this.


kadevgraaf:
Thanks. It works quite intuitive now.

PR #347, opened 2017-04-04
Swik 1123 apply themes to deck themes tester

huwf:
Part of SWIK-1123This applies as CSS theme to a slide in slide view and edit mode, loading on the client in componentDidMount.  In addition, it modifies the content route, adding an optional theme parameter.  

This can be used to preview a slide with an arbitrary theme not assigned to it through a service.  The theme name must be one of the reveal.js themes.  It does not check for incorrect strings, but does for empty and undefined, in which case the default white theme is applied


components/Deck/ContentPanel/DeckModes/DeckEditPanel/DeckPropertiesEditor.js
```diff
@@ -392,15 +392,15 @@ class DeckPropertiesEditor extends React.Component {
             <option value="default">Default - Reveal.js White</option>
             <option value="beige">Reveal.js Beige</option>
             <option value="black">Reveal.js Black</option>
-            <option value="blood">Reveal.js Blood</option>
```
kadevgraaf: I commented these because they are not yet supported in reveal.js

stores/ContentStore.js
```diff
@@ -3,11 +3,18 @@ import {BaseStore} from 'fluxible/addons';
 class ContentStore extends BaseStore {
     constructor(dispatcher) {
         super(dispatcher);
-        this.selector = {'id': 0, 'spath': '', 'sid': 0, 'stype': '', page: 'content'};
+        this.selector = {'id': 0, 'spath': '', 'sid': 0, 'stype': '', page: 'content', 'theme': 'white'};
```
kadevgraaf: Does this make the theme hardcoded?

services/presentation.js
```diff
@@ -16,10 +16,10 @@ export default {
             /*********connect to microservices*************/
             let returnErr = false;
             let slideServiceRes;
-            let theme = get_sample_theme();
+            //let theme = get_sample_theme();
             rp.get({uri: Microservices.deck.uri + '/deck/' + String(args.id) + '/slides'}).then((res) => {
                 slideServiceRes = JSON.parse(res);
-                callback(null, {content: slideServiceRes.children, theme: theme, selector: selector});
+                callback(null, {content: slideServiceRes.children, theme: slideServiceRes.theme, selector: selector});
```
kadevgraaf: I added this to make use of an update to the /slides route, which now also returns the theme from deck-service

PR #353, opened 2017-04-06
Swik 1176 separate slide title from header in content

dpaun:
This, very small, change separates slide's title from CKeditor's inlineHeader. Currently, when a slide is edited, its title receives a value from inlineHeader, which is wrapped in <p> tags. This is not visible in the decktree, because the deck-service uses striptags module before returning a decktree. When this change is applied, editing a slide will not change slide's title anymore.


kadevgraaf:
Thanks Dejan!

PR #354, opened 2017-04-06
Swik 1160 allow visiting existing fork in permissions modal

stavmars:


kadevgraaf:
Works well!
Next to a link in the text, a more explicit large button (next to the others) might be an additional improvement. If people fork less, it'll save space (and they wont be confused by many forks), so we have to encourage them I think.

PR #356, opened 2017-04-07
Swik 216 like feature

ntomasevic:
Implementation of like feature. 
“Like” feature is deployed per deck, and it is allowed only for users who are logged in (disabled for unregistered users). It allows users to like a deck, and to withdraw liked activity already performed. It also includes an indicator of total number of likes per deck (visualized at deck view).


kadevgraaf:
Works nicely 👍 
I am ok with this being merged. 
A small optimization would be to check for multiple likes of a deck by a single user.
Now you can like a deck on multiple slides (next to deckviewpage), wich would then result in the screenshot below for e.g. deck http://localhost:3000/deck/1255-1.
screen shot 2017-04-07 at 11 49 17

Sole-Valero:
I can't test it, experimental doesn't works...so I can't run any branch

ntomasevic:
Issue related to multiple likes from different slides within a deck is now resolved. Ready for merging :)

PR #357, opened 2017-04-07
Swik 1046 adding subdeck

Sole-Valero:
This is the first version of the modal which allows to add a deck as a subdeck. It gives the user two options: select it from their own slides and from SlideWiki. If the users select from SlideWiki, it first has a small list with the most recent decks..but it can search for the deck it wants in a similar way it does in the search panel of the platform.
You can access the new button from the toolbar of the slide or from the toolbar from the treenode. If you have permission to add a deck...the button is enabled. If no, it is shown disabled.

Also available in my test server: http://svalero.dsic.upv.es:3000/deck/1239-1


abijames:
I've tested UI and ok for this to merge

mmeimaris:
After merging the latest master into this branch, it doesn't seem to be working. I'm getting the error 
`2017-04-07T14:27:30.803Z - error: uncaughtException: Cannot read property 'permissions' of undefined date=Fri Apr 07 2017 17:27:30 GMT+0300 (GTB Daylight Time), pid=7912, uid=null, gid=null, cwd=C:\Users\Marios\Documents\Projects\SlideWiki\original\org3\slidewiki-platform, execPath=C:\Program Files\nodejs\node.exe, version=v6.3.0, argv=[C:\Program Files\nodejs\node.exe, C:\Users\Marios\Documents\Projects\SlideWiki\original\org3\slidewiki-platform\start.js], rss=299212800, heapTotal=296833024, heapUsed=162798576, loadavg=[0, 0, 0], uptime=795870.3013051
TypeError: Cannot read property 'permissions' of undefined
    at TreeNode.render (C:/Users/Marios/Documents/Projects/SlideWiki/original/org3/slidewiki-platform/components/Deck/TreePanel/TreeNode.js:154:31)
`

@Sole-Valero @stavmars any insight on this? Have you tested this with the new edit rights that are now merged into master?

Prior to the merge I was working on this branch without problems (a few hours ago).

stavmars:
Yes, the problem is in line TreeNode.js:154:31. It tries to access the undefined property PermissionsStore instead of directly accessing this.props.permissions.  @Sole-Valero do you want me to fix this?

kadevgraaf:
Thanks. Works perfect now afaik. Very usable and intuitive feature! 👍

PR #369, opened 2017-04-13
Swik 1134 forbid editing old deck revisions

stavmars:
This PR addresses issues SWIK-1134 and SWIK-1200. 
It forbids editing older root deck revisions and shows appropriate modal if someone is an editor and tries to edit a read only revision. For people without edit rights, the modal just informs them about ways to gain edit access. Also, a bug is solved when going back through browser history the no permissions modal appears again.


kadevgraaf:
Works well!

PR #370, opened 2017-04-13
Swik 1223 removing duplicate service error msg from site

vinaymodi:
Please test at your end. I have tested by changing user microservice url. Without using this solution, it shows two error messages but after using this solution there is only one message.


kadevgraaf:
Works well. Thanks!

PR #373, opened 2017-04-18
Fix bug when clicking edit tab without edit permissions

stavmars:


kadevgraaf:
Seems to work well. Thanks for the fix!

PR #374, opened 2017-04-18
Swik-1123 apply themes to deck themes tester

huwf:
This PR requires that the SWIK-983 change in the reveal.js submodule also be complete.

Main change is that we don't return white.css as a theme if we have a theme with "default".  The white.css file should only be used for legacy, and default.css should be used for the new themes.

It also adds in all themes for the addDeck component, since they should all be ready for deployment. 

A bug with the display in view and possibly edit mode is highlighted with this change, particularly on e.g., serif, but this existed anyway, and came as a result of changing the .reveal div to position:absolute.  I'll create a separate ticket for that.


kadevgraaf:
Works well. Thanks!

PR #376, opened 2017-04-20
Swik 1216 prevent undefined values in activitiesservice calls

dpaun:
This fix solves the problem with undefined values in activity-service calls.
Content_kind and content_id are validated before sending a request to the activity-service.
Also, links in ActivityFeedPanel and ActivityItem are modified in case of invalid data.


kadevgraaf:
Works well. 
Redirect from addeck still seems to render incompletely.

PR #380, opened 2017-04-21
Swik 1234 inconsistent views related deck

TBoonX:
I just enhanced the general LanguageDropdown component, used it in Edit Deck component, reordered one field and used full names for licences in Edit Deck.


kadevgraaf:
License change looks good!
The language however changes back to the original language when saving the deck. No worries though - I think it is still fine if we include this in the next release.

abijames:
Some accessibility issues have crept into this form. Can you add htmlFor on the language label and an ID on the input box?
I will make a new task for some other issues on this form so if this is not going to be merged immediately they may be easier to do all together.

TBoonX:
@abijames I did what you asked for.
@kadevgraaf the language dropdown does not work as expected - I will fix it

TBoonX:
@kadevgraaf please check again

PR #382, opened 2017-04-21
Fix deck pages with revision on URL not loading after forkCount fix

kprist:


kadevgraaf:
Works well. Thanks for the quick fix!

PR #383, opened 2017-04-24
Fix permissions modal showing again after fork

stavmars:


abijames:
Working correctly. Ok to merge

PR #391, opened 2017-05-09
Put back missing registration of like service

dpaun:
Putting back the missing line of code which registers like service.


bwulff:
odd, when did the line go away? or was it not merged in the first place?

dpaun:
I'm not sure. Maybe when one of our latest commits was merged. 
Swik 216 like feature #356 was merged, 29 days ago. It's working on stable.

PR #393, opened 2017-05-10
842 implement reporting function/button/dialog

sednanref:
Report modal now available for not logged-in users, in this case, the modal asks for a name and a Captcha validation.


kadevgraaf:
Works well. Nice Recaptcha!

PR #394, opened 2017-05-11
Swik 1308 mark deck items

Sole-Valero:
Now, when an user selects one deck from the list, is is marked with the aria-label   aria-selected=true. For items not selected,aria-selected=false
Branch SWIK-1308-markDeckItems
Available at. http://svalero.dsic.upv.es:3000/


abijames:
ok to merge

PR #397, opened 2017-05-15
Swik 1326 download and fork activities

dpaun:
Download (epub and pdf) and fork activities are recorded using the activity-service (and shown in the activity feed panel).
Fork activity_type is added, since it wasn't a part of the initial implementation.
Adding of 'edit' activity when renaming a node was implemented.
Also some modifications were made to handle unknown user_id values.


kadevgraaf:
Works good!

PR #398, opened 2017-05-17
Swik 568 refine progress bar

dpaun:
This branch improves L&F of the import progress bar.
While uploading a file, the progress bar percentage is increased every 2s (reintroducing Kurt's code, with some modifications), until 65%, for the amount which is calculated based on the file size. Since currently we don't have the real information about the file upload progress, this calculation is an estimation and will be more or less accurate depending on the user's network upload speed. If, after reaching 65%, the upload is not yet finished, the message will change from 'Uploading file' to 'Converting file'.
Accessibility is also improved.

### DIFF
--- a/actions/import/uploadFile.js
+++ b/actions/import/uploadFile.js
@@ -14,7 +14,7 @@ export default function uploadFile(context, payload, done) {
     const timeout = 60;
     const remainingProgress = 55;
     let currentProgress = 0;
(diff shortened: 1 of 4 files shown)


dpaun:
Thanks! I have included your suggestions.

abijames:
@bwulff can you merge this?

PR #399, opened 2017-05-17
SWIQ-1100_hashing_password_on_server

TBoonX:
Outcommented API key in configs/general.js. Adopted changes of resetPassword route.


rmeissn:
Review: works like a charm, looks good.

PR #404, opened 2017-05-25
Swik 1239 fix activity feed width

dpaun:
Fixing Activity feed panel layout. Minor changes - some styles were modified to fix panel width.
Also fixed SWIK-1198 'Activity feed shows only the bottom element after hiding the decktree', by changing ReactList type to 'simple'.


kadevgraaf:
Works well!

PR #405, opened 2017-05-25
Testing framework setup

txwkx:
- install `Enzyme` and `Jest`
- modify test environment *test/setup.js*
- add UserPicture example tests


components/common/UserPicture.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
```
rmeissn: Isn't it save to remove "React" as an import, as it's not used anymore in the entire file?

package.json
```diff
@@ -73,7 +73,7 @@
     "fluxible-addons-react": "^0.2.9",
     "fluxible-plugin-fetchr": "^0.3.11",
     "fluxible-router": "^0.4.21",
-    "focus-trap-react": "^3.0.2",
+    "focus-trap-react": "^3.0.1",
```
rmeissn: Why did you decreased the version number?

package.json
```diff
@@ -33,7 +33,7 @@
     "dev:windows:dashboard": "set HOST=127.0.0.1&& set PORT=3000&& set DASHBOARD=1&& webpack-dashboard -- webpack-dev-server --config ./webpack/dev-server.js",
     "dev:windows": "set HOST=127.0.0.1&& set PORT=3000&& node ./webpack/dev-server.js",
     "lint": "eslint --ignore-path .gitignore -c .eslintrc \"./**/*.js\"",
-    "test": "mocha test/unit/ --recursive --compilers js:babel-register --reporter spec",
```
rmeissn: You've outcommented unit tests by this change. Please reinclude them or introduce different tests levels (e.g. test:components, test:unit, test:...).

components/common/UserPicture.js
txwkx: No, it's not safe, would error `React is not defined`.
Because JSX wants React to be in scope.

txwkx:
@rmeissn 
1. No it's not necessary, I followed the common naming convention for tests which is *.test / *.spec / *-test / *-spec.


2. Those are the only current out of the box options (error comments). 
```
expected *** to have a length of 1 but got 0
```
```
expect(wrapper.find(***)).to.have.length(1);
```
But yeah, sometimes they aren't as obvious and straightforward

package.json
txwkx: updated +

package.json
txwkx: `git rebase` shenanigans

components/common/UserPicture.js
rmeissn: Okay

rmeissn:
Alright. As you modified the test commands of the file `package.json`, please also change the Travis execution of tests to include all kinds of them and upload of the coverage report to Coveralls (also Travis). Make sure that the coverage report includes all test categories. See line 38 in `package.json` for the coverage report generation.

rmeissn:
Can you please update this branch with latest commits from master and resolve the conflicts? I like to review it as of this sprint again.

txwkx:
@rmeissn 
- rebased the branch
- updated enzyme to v3
- linked with coveralls service

.travis.yml
```diff
@@ -21,6 +21,7 @@ after_success:
 - openssl aes-256-cbc -K $encrypted_ecdba13b2a41_key -iv $encrypted_ecdba13b2a41_iv -in deployment_keys.tar.enc -out deployment_keys.tar -d
```
rmeissn: Please introduce different lines for the different tests (unit, components, ...) instead of just executing `npm run test`

package.json
```diff
@@ -33,9 +33,10 @@
     "dev:windows": "set HOST=127.0.0.1&& set PORT=3000&& node ./webpack/dev-server.js",
     "langs": "babel plugins/intl/translate.js | node",
     "lint": "eslint --ignore-path .gitignore -c .eslintrc \"./**/*.js\"",
-    "test": "mocha test/unit/ --recursive --compilers js:babel-register --reporter spec",
```
rmeissn: You've deleted the `test` command but you're using it int the travis.yml

package.json
```diff
@@ -155,20 +156,23 @@
...
+    "enzyme": "3.1.0",
+    "enzyme-adapter-react-15": "1.0.2",
     "es6-promise": "^4.1.0",
     "eslint": "^3.17.0",
     "eslint-plugin-babel": "^4.1.1",
     "eslint-plugin-react": "^6.10.0",
     "extract-text-webpack-plugin": "^2.0.0",
     "husky": "^0.13.2",
-    "istanbul": "0.4.5",
+    "istanbul": "1.0.0-alpha.2",
```
rmeissn: Why do you use an alpha version?

.travis.yml
txwkx: The "unit" folder contains:
- an empty test file
- a trial test file with some default values

In my opinion having a separate `npm:unit` command is currently irrelevant. Could be added later when "unit" test are written.

package.json
txwkx: This issue: https://github.com/gotwarlost/istanbul/issues/262
Although the coverage results still seem to be inaccurate

.travis.yml
rmeissn: We do not need to run unit-tests as of Travis but I think it's useful to have a unit test command in the package.json, even though there are no useful tests yet. Otherwise people might come up with a different technology or package for unit-tests than what we planned to use.

package.json
rmeissn: Definitely. If it stays this inaccurate, it's better to deactivate coverage generation and upload to coveralls as it will result in a false impression about the project.

PR #406, opened 2017-05-25
Fix usage tab not showing users and titles

stavmars:
This solves a bug where the usage tab did not show on some occasions titles and usernames.


kadevgraaf:
👍

PR #407, opened 2017-05-26
Swik 579 implement share button

dpaun:
This PR introduces sharing of decks and slides on social networks (Twitter, Facebook, Google+, LinkedIn) and by email.


dpaun:
One note related to localhost testing: Sending 'localhost' urls to Facebook and Google fails because it seems that these two are checking whether the links are valid.

kadevgraaf:
Works well!

Facebook sharing is not currently supported - "URL is not added to domain of Facebook app", so I think we probably need to disable it for now (@abijames - I recall Facebook login was also disabled for this reason, right?)

The activity feed correctly shows a notification when you share something. However, the notification in activity feed is created before sharing is completed (you can cancel a share-activity, but the notification remains). Would it be easy to have a feedback mechanism to check if a share-activity was completed, and only then add the share-activity to the activity feed? No worries if this is difficult or time-consuming to implement: we can skip this for now and create a JIRA task (with low priority I think) until after the release. We might even include the result of a share-activity (e.g., link to twitter feed) using this feedback-mechanism.

dpaun:
I agree that having a new notification in the activity feed before sharing is completed is a bit of a problem. I haven't yet found a solution for this. A new task which would handle this seems like a good idea.

kadevgraaf:
Hi @dpaun !
Great. I did not see your comment "Sending 'localhost' urls to Facebook and Google fails because it seems that these two are checking whether the links are valid." at the time, so then Facebook would probably be OK if it's on the server.

abijames:
Works well and adds lots of nice functionality for users. There are some accessibility issues that need to be fixed but I would recommend we merge this PR and I'll put notes on the ticket as we may want to make it new task.

I also recall that there was a discussion about permanent URLs for decks.... https://slidewiki.atlassian.net/wiki/display/SWIK/Permanent+Deck+URLs .... it would be good to use these when sharing decks if we implement it

PR #408, opened 2017-05-31
Swik 1280 username case insensitive

TBoonX:
I just changed the regex for usernames when signing up or in.
Should be merged after the same branch in the user-service got merged.


kadevgraaf:
Works well. 
I registered with username Kadegraaf and email kgf300@vu.nl
later I registered with username KAdegraaf and email KGF300@vu.nl and I got an error that username was taken. Username is now also uppercase

PR #409, opened 2017-05-31
SWIK-1007 Added jwt to the payload - A very small change

dpaun:
A very small change:
Added jwt to the import service call payload, since this is required by the new file-service API which the import-service will be using in the future when processing images.
This is required before we merge the import-service branch.


kadevgraaf:
Thanks!

PR #421, opened 2017-06-06
SWIK-1304_userservice_404_breaks_flux

TBoonX:
I just added wrong data handling in the action fetchUser which fixes the unwanted page reload and 404 Error


kadevgraaf:
Works well!

PR #423, opened 2017-06-06
SWIK 644 Display notifications for users decks slides

kadevgraaf:


abijames:
Comments on https://slidewiki.atlassian.net/browse/SWIK-644
If the keyboard accessibility is improved then I think this is ok to merge and then improve on later

kadevgraaf:
Looks good!

PR #425, opened 2017-06-06
Swik 1356 append slides

Sole-Valero:
It adds a button to the node tool bar and another to the content header toolbar which open a modal, This modal allows users to attach the slides they select. 
The slides will be attach in the order the user selects them. 
It is a preliminar version with some accesibility issues about navigate throug deck list and slide grid, because user can't use arrow keys and only tab. 
Furthermore, it improves the attachDeckModal regargind some accebility issues.
Original attachDeckModal was refactored into more small components. These components are use now for both modals. 
Finally, the seach panel was updated to use only react-semantic-ui elements.
This branch is also accesible from my server: http://svalero.dsic.upv.es:3000


Sole-Valero:
@kprist is fixing some issue about how the service finally attach the slides.

Sole-Valero:
@kprist has solved the issue

kprist:
Hi there are still issues with this, when attaching multiple slides in a subdeck (not a deck at the root). At first it appears to work, and they appear *inside* the subdeck. But the service handles the request differently and adds then *after* the subdeck, as siblings to it. So if you reload the deck page you'll see them in different position.
I am changing the deck service in order to handle that the same way the platform assumes it will be handled.

kprist:
Deck service has been fixed for known issues regarding attaching slides. Also fixed thumbnail generation not working.

PR #431, opened 2017-06-07
Swik-1398-route that handles legacy URLs (also disabled SWIK-1287-translation-UI code)

DaryaTarasowa:
redirects


kadevgraaf:
This is about two tasks - 
Swik-1398-route that handles legacy URLs - @bwulff and @kprist  were much involved in this if I recall correctly.
and SWIK-1287-translation-UI - but it is disabled - this part seems safe to merge as it is disabled... SWIK-1398 still needs to be checked though

kadevgraaf:
Made https://github.com/slidewiki/slidewiki-platform/pull/440 without translation code

PR #434, opened 2017-06-08
Swik 644 display notifications for users decks slides

dpaun:
Small modification to improve (fix) marking notifications as read
Also modified display of semantic ui 'reply' icon which for some reason is smaller then other.


txwkx:
[off topic]
Apparently, clicking on a username does not go to the profile page, but requests a review in a PR you have nothing to do with. 
 `¯\_(ツ)_/¯`

PR #435, opened 2017-06-08
SWIK-1346_userprofile_could_not_be_saved

TBoonX:
save user profile now reads language from cookie

Have to be merged with same branch of user-service


kadevgraaf:
Works well. Thanks!

PR #436, opened 2017-06-08
Fix error when clicking view tab

stavmars:


kadevgraaf:
Bug fixed. Many thanks!

PR #441, opened 2017-06-13
Added check for duplicate existing IDs and for duplicate random IDs in slide content

kadevgraaf:


txwkx:
Looks good 👍 :shipit:

PR #443, opened 2017-06-14
replace revision count with fork count in recent decks listing

schatzopoulos:
a PR in deck-service needs to be merged before this


kadevgraaf:
Works well!

PR #445, opened 2017-06-14
SWIK-1411_language_is_read_wrong_in_userprofile_and_registration

TBoonX:
Added reading language from cookie to common.js and used it in the components which send the language to the userservice


rmeissn:
Looks good. @kadevgraaf is it save to merge this fix now or do we wait for something?

PR #449, opened 2017-06-14
Swik 1428 fix activity feed for slides with special letters in titles

dpaun:
Activity feed panel (Data sources and Notifications panels, also) doesn't properly display special letters (like Spanish Accents) in slide titles. This has been solved in this branch, using cheerio.


kadevgraaf:
Works well!

PR #453, opened 2017-06-15
Swik 1429: autodisable the action button 

Sole-Valero:
Action button are disabled when the user enters into the edit mode.
In order to check,  select a deck with edit rights, then enter and exit from edit mode and see the results.


abijames:
The deactivated buttons need to be removed from the tab order by setting tabindex=-1

Sole-Valero:
Problem solved.

abijames:
@kadevgraaf this is tested and you can merge it

PR #455, opened 2017-06-15
Swik 1431 create deckfamily page

schatzopoulos:


kadevgraaf:
Works well. Tested with http://localhost:3000//deckfamily/test (3 decks with tag "test" returned) and http://localhost:3000//deckfamily/test1 (1 deck with tag "test1" returned)

PR #457, opened 2017-06-15
Swik 1163 security issues

TBoonX:
I moved the secrets to secrets.js which is just included into the services. Thus the secrets are hidden in the client. I also added that the secrets and the public recaptcha key could be changed via env.


dpaun:
Looks good!

PR #458, opened 2017-06-15
SWIK 1339 info panel component

kadevgraaf:


kadevgraaf:
I can also easily relatively easily show both the deck and slide, or the entire breadcrumb in the infopanel, so we can decide to include this all and remove the breadcrum (title is now shown 3 times on a page)

PR #460, opened 2017-06-16
Swik 1441 add images license checkbox to add deck

dpaun:
Added a checkbox to the 'Add deck' dialog for the user to agree on licenses of imported images.

PR #461, opened 2017-06-16
SWIK-1444 Uploading file and creating a new deck on 'Add Deck' button…

dpaun:
The import pptx process has been modified.
With this PR, a pptx/odp file is uploaded and the deck created on 'Add deck' button action, after the user has agreed on terms and conditions.


kadevgraaf:
Looks good w.r.t. flow and functionality!

But same here as in https://github.com/slidewiki/slidewiki-platform/pull/460: getting the following error in browser console ( not in terminal though :) ) : 
Uncaught Error: NavLink created with empty or missing href ''or unresolvable routeName 'undefined

kadevgraaf:
Ah sorry. It does work, despite some errors/warnings.

PR #463, opened 2017-06-16
Swik 1455 unrequire comment text

dpaun:
This is a small fix which allows adding comment (and replies) without the text of the comment. The title field is still required.
Required changes on the discussion-service side are already implemented.


abijames:
Works well! @kadevgraaf happy to merge

PR #464, opened 2017-06-16
SWIK-1452 move language chooser

rmeissn:
* Moved language chooser to user-settings (right column)
* Rename to "User Interface Language"
* Deactivate auto update feature
* Refactored LocaleSwitcher component


kadevgraaf:
Works well!

PR #466, opened 2017-06-18
Swik 1453 501 after deck creation

stavmars:
This PR addresses issues [SWIK-1465](https://slidewiki.atlassian.net/browse/SWIK-1465) and [SWIK-1453](https://slidewiki.atlassian.net/browse/SWIK-1453).


kadevgraaf:
Works well. Thanks for the fix!

PR #470, opened 2017-06-20
Created license info page, added link to footer; improved layout About page

abijames:
This adds a page describing the Creative Common license used in SW. Available at /license 
It also merges the updated footer layout.


abijames:
@kadevgraaf @ben this branch doesn't see to have merged correctly. It should have updated the footer and there should now be a page https://platform.experimental.slidewiki.org/license

PR #472, opened 2017-06-21
Swik 1446 home page - new design including carousel  

abijames:
Content on the page will still need completing but this adds in a new library so we need to merge for FFF


ali1k:
I think this PR should go to the public release, however the following improvements still need to get in there:
- add appropriate images for main features
- add title of main features
- add a short description for features in the carousel
- find a better place for the 'pause' button.
- UI internationalization for home page

abijames:
@ali1k I agree with your suggests and @bwulff is happy for content/layout changes to happen on the home page after the feature freeze. We are just keen to merge it into master now so that the changes to the libraries can be included in testing and link to new pages can be included.

ali1k:
@abijames OK. I just added some sample feature titles and changed the position of 'pause' button a bit!

PR #473, opened 2017-06-21
Swik 1449 refine tag panel

schatzopoulos:
- Adds autocomplete functionality when inserting tags
- Tag Panel from slides needs to be removed


abijames:
I am generally happy with the UI but there needs to be some accessibility fixes.

At the moment, if you click on a tab the deckfamily page opens but does not include the current deck. @kprist is this the issue you were commenting?

schatzopoulos:
@abijames this definitely shouldn't happen. Probably this is an issue with the synchronization of the search index on testing. I have to check it with Ben but I remember fixing it during the hackathon. Maybe something has changed in the deployment.

schatzopoulos:
@abijames I added some aria-labels in tags and button in the tags panel. Please, inform me if further accessibility issues need to be resolved.

The issue about deckfamily seems to be working now, so it was probably something in the server deployment that has been fixed.

abijames:
ok to merge. Some accessibility issues still to deal with but functionality is a good improvement @kadevgraaf

PR #474, opened 2017-06-21
Improve-travis-built-time 

rmeissn:
I've implemented a more efficient version for installing NPM packages. They are now only updated (or downloaded if cache is missing) and are not downloaded a second time as part of the docker build execution. The resulting image content stays the same.
By updating the packages as part of the Travis run, we also gain more security as the software will be tested against the exact same packages as installed in the docker image.

Please answer the comment I've added to package.json before merging this:
Why are we executing "npm install" instead of "npm install --production"? I'd like to exchange it with "npm prune --production". Do we need the dev dependencies at our production deployment?

What is "npm run install" doing? Do we need it? It always responds with "napa".


Dockerfile
```diff
@@ -10,6 +10,7 @@ WORKDIR /nodeApp
 ADD . /nodeApp
 RUN ./make_version.sh
 
+# why is this not "npm install --production"? I'd like to exchange it with "npm prune --production"
```
kprist: the reason we have two lines relates to the `napa` (https://github.com/shama/napa) dev module that imports the line-height plugin as a node module (it is not). That generally runs correctly with just `npm install`, but not when running as root (as is the case in deployment) because of https://github.com/npm/npm/issues/5637. Running `npm run install` overcomes that.

With your new approach I would guess that you could replace it with `npm prune --production` but still keep the `npm run install` command. That only runs `napa` and fetches the line-height plugin, which is a runtime-dependency.

kprist: actually, I just tested my own suggestion and it breaks things in that you can't `npm run install` after pruning, because napa ( a dev dependency ) is itself removed. Pruning also removes the (virtual, unknown module `ckeditor-lineheight-plugin` that napa installs. This is too complicated just for a ckeditor plugin!

Anyway, I think the correct way might be to add `npm run install` **after** `npm update --dev`, and then replace these two lines with:

```
npm prune --production
npm install napa
npm run install
```
It should work for now, but maybe we should rethink about how we import this one ckeditor plugin

rmeissn: Sounds well thought and I can't see why it shouldn't work.

@anyone else: Any other reasons for keeping dev deps as part of the production environment?

PR #475, opened 2017-06-22
Swik 1007 use jwt when uploading images

dpaun:
This PR replaces usages of userId parameter, in CKEditor image uploading configuration, with jwt token. The first implementation of image uploading functionality was implemented using a shared folder and this required the usage of userId. The new implementation (already on import-service master, backward compatible) uses the new file-service API which requires the jwt parameter when uploading images.


dpaun:
Yes, I think this is true.
I will close the PR for now and make necessary changes.

PR #476, opened 2017-06-22
SWIK-1447-MoveAcrossGrid user can move acros the grid of slides using…

Sole-Valero:
This change allow users to navigate throug the grid using arrowKeys or tab. For selecting/deselecting a slide, users should continue using enter key or mouse. 

Ticket: https://slidewiki.atlassian.net/browse/SWIK-1447


kadevgraaf:
@abijames  - Should this go in the release as a fix, or is it a new feature? (ticket says: improvement :) )

abijames:
@Sole-Valero @kadevgraaf I suggest this does not go into the release. There still needs to be some improvements on the accessibility of the modals as we seem to have broken the dialog description and I was experiencing focus issues with the grid.

Sole-Valero:
@abijames I didn't touch the dialog description...what is the problem you have? 
About the focus issues. what is the matter?

bwulff:
There is a conflict here due to which the branch cannot be merged. 
Accidentally  commited somehting from the custom-semantic-ui submodule?

Sole-Valero:
@bwulff I merged it with master four days ago...I didn't touch anything from custom-semantic... how we can solve the issue?
But...also @abijames commented something about don't merge it....

Sole-Valero:
@abijames I have move the modal explanation to the end of the content panel  in the second screen of the attach slides modal. 
We can't send the focus to an element which is not accessible by tab, so the only way I found to allow users to listen the modal explanation is moving it until the attach and cancel buttons.

Sole-Valero:
he modal to attach slides has been patched:
in the first screen, users can achieve the list of decks with tab key. Then can move up and down with arrow keys. If they press again tab key, they go to the next button.
in second screen, the focus goes to the label which contains the deck selected, in order to allow sw readers to pass through the modal explanation. If users press 'all slides button' or select by hand all the slides, the focus goes to the attach button. Also, users can move by the grid using arrow keys. If the press tab, they go out the list to an action button.
Also, as the deck list component is shared with the attach deck modal, now also in this modal the list is fully key accessible.

abijames:
@kadevgraaf as suggested by Sole we suggest merging this now and we will fix further accessibility issues later

PR #477, opened 2017-06-22
Fixed accessibility. Improved text layout

abijames:
Improves accessibility on the deckview page and improves the layout of the text. Also introduces region/landmark element (main) to assist with navigating content.


kadevgraaf:
Should this go in the release as a fix?

abijames:
If possible it should go into the release. There will be other fixes like this to come but they will not require much testing (I'll do it).

kadevgraaf:
Looks good!

PR #478, opened 2017-06-23
Swik 1484 add greek ui language

kprist:


abijames:
@kadevgraaf I happy for this to be merged.

PR #479, opened 2017-06-23
header searchbox fix on icon click

schatzopoulos:
Header search box fix when clicking the search icon next to the input


kadevgraaf:
Works well!

PR #481, opened 2017-06-26
SWIK-1472 Fix problem with content name in the activity feed

dpaun:
A very small fix for the problem with undefined content name in the activity feed (identified during like functionality testing).


kadevgraaf:
Works well!

PR #484, opened 2017-06-27
SWIK-1498 Bugfix for sources panel and wrong content modules type

dpaun:
A small bugfix for the problem with undefined title in sources and notifications panels. Also commented the line which was setting module type to 'contributors', since contributors currently is not one of the content module tabs (this caused the comments tab content to be displayed by default).


kadevgraaf:
Proper fix. Thanks!

PR #487, opened 2017-06-27
Swik 1426 scorm button

Sole-Valero:
Now, the download button is a Dropdown reac-semantic-ui component which allows users to select the format desired and download the file.

@abijames should check the accesibility of the feature.


Sole-Valero:
Now the button is accesible by keyboard. I should remove the header of the menu. This element form the react-semantic-ui dropdwn causes that you can not navigate through items in the list. the only way is to provide the options using the prop options. the other ways are still under construction.
Perhaps we should think about using the semantic ui element for this button.
Furthermore, there are some warnings errors than dissapear in production mode (built). They appear because I forced to render the icon as "download large". Another time, in the component now it is only allowd to pass the icon you use as a prop, an only expects "download". But if we use only the name, the icon is rendered so small. It is not allow adding Icon as a child element ...

Sole-Valero:
@abijames , @bwulff , @kadevgraaf : Abi detected a problem with screen reader: they don't read aloud the options. The only way to solve the problem is modifying the main project react-semantic-ui. I'm trying they accept my update. We have the same matter in other dropdwons...so perhaps we can add now the funcionality as it is.

abijames:
@Sole-Valero do you suggest we merge this now so that at least the button is working with the keyboard?

Sole-Valero:
Yes @abijames, I solved it the react-semantic-ui project, but I'm still waiting they merge the solution. When they do, we only need to get the last version of the module and the problem will be solved.

Sole-Valero:
Now the last version of react-semantic-ui is used. This version fixes the problem of the dropdoown button with readers.

Sole-Valero:
Last version of react-semantic-ui introduces our changes...but also other one in the dropDown than causes our logic don't work. So, we need to revert this branch two commits before...to the one e0c0c87. Then we can merge it and solve the new issues in following versions.

kadevgraaf:
I get a warning that 'download large' is not correct for semantic-ui-react (the 'large' tag). "large" does seem to work though, so for now we merge this.

Sole-Valero:
Only in dev mode...in production no

PR #491, opened 2017-06-28
Removed logging of error object

vinaymodi:
Please verify and close as needed.


bwulff:
Thanks! That already removes a lot of clutter.
I find the request that is still reported in the error messages quite useful but unfortunately they clutter the log too. Let's see how we can extract useful information from it in another story.

PR #493, opened 2017-06-29
Swik 1324 show deck forks as deck other versions

schatzopoulos:
This shows deck forks instead of deck revisions as deck "other versions" in search results


bwulff:
Thanks, Serfeim!

PR #494, opened 2017-06-30
SWIK-1479 SWIK-1475 allow restore to original slide version, disable for active version

kprist:


kadevgraaf:
This fixes the bug w.r.t. restoring original slide - https://slidewiki.atlassian.net/browse/SWIK-1479 , so a very nice improvement.
Will merge, however, it not intuitive for users how you get back to the latest slide revision. 
The messages/notifications say: "restored to earlier version", but you can also restore the latest, or a future version (depending from where you restore from). 
Perhaps a mouse-over tooltip would be good with instructions + a preview to show what you are restoring (there is now the view button, but this takes you to a different tab). Now the restore button is not clear on whether you "undo" a version or whether you "go to" (restore) this specific version. Created SWIK task https://slidewiki.atlassian.net/browse/SWIK-1552 for above

PR #500, opened 2017-07-07
Swik 1527-bugfix  alert to confirm slide/subdeck deletion

Sole-Valero:
When user press on the delete button (from both toolbars), a sweet alert is shown to allow user confirm the action.


abijames:

PR #504, opened 2017-07-10
Make html id unique even for duplicate slides in presentation mode

stavmars:


Sole-Valero:
I'v tested it and I think it works perfectly.
I created a deck to test it (2156): This deck has the same deck attached two times. Also, the same slide attached two times. Now, all the elements are displayed in the presentation mode.

kadevgraaf:
Works well. Thanks!

PR #506, opened 2017-07-11
Navigate to next slide/subdeck when deleting a node

stavmars:


abijames:
@kadevgraaf  this is ok to merge

PR #508, opened 2017-07-13
Swik 1317 loading indicator reset password

TBoonX:


TBoonX:
I added a loading indicator. I used the form loading indicator.

PR #509, opened 2017-07-13
Swik 1521 bug

DaryaTarasowa:
fix, the architecture is slightly changed


kadevgraaf:
Tested on my localhost by connecting to testing server microservices in configs/microservices.js (since experimental does not have the original SW.org decks)

http://localhost:3000/deck/345_information-retrieval-lecture-series#tree-0-deck-345-1-view redirects to http://localhost:3000/deck/220**-1#tree-0-deck-345-1-view** 

The latter part does not seem to prevent basic view functionality, and at least it is shorter/more accurate than previous URL.

PR #512, opened 2017-07-14
SWIK-1539_listed_group_creato

TBoonX:
Added creator of a group as member to member counts and group edit view. Also improved a few texts.
I also assured that the group details modal, could be opened in deck edit view, is rendered correctly by showing it after the data was getted.


abijames:
This is ok to merge. 

But have one suggestion... you could replace "You are the creator, which makes you also a member" with "Group owner" to make it simpler

TBoonX:
@abijames Which option do you mean/want:
1. "You are the group owner, which makes you also a member"
2. "Group owner"

abijames:
2. Group owner

TBoonX:
I asked because it seems to be very unclear to users that they are the creator and member at once.

PR #513, opened 2017-07-14
SWIK 1535 ImagesFeatureText

kadevgraaf:


abijames:
@ali1k @kadevgraaf I have updated this branch with larger images for the carousel, added text to the menu on the carousel and improved the layout of other information on the home page. I feel this is now ready to merge into Master as we need to get it into the release.

ali1k:
my comments:
- links to the features page need to be fixed. And currently there is no content in that page.
- I think we can still use larger and more descriptive images to convey our message.

in my opinion, it is good enough to be merged into master. Later on we can improve it further...

abijames:
@ali1k I am working on the Features page today (plus /welcome that is needed as well)

PR #522, opened 2017-07-19
SWIK-1213_remove_facebook

TBoonX:
removed facebook from Integrations component


TBoonX:
Should go into the release

PR #530, opened 2017-08-03
Swik 1588 add emailing functionality

TBoonX:
I added a new service _email_
With calling create on it it will try to send an email like the user-service. Running the platform with _npm run dev_ makes the service return success without connecting to the SMTP server.
This could not be tested completely because experimental has no SMTP server


package.json
```diff
@@ -28,7 +28,7 @@
     "build:windows": "webpack --config ./webpack/prod.config.js && set NODE_ENV=production && npm start",
     "build:nostart": "webpack --config ./webpack/prod.config.js",
     "build:langs": "babel plugins/intl/translate.js | node",
-    "dev": "node ./webpack/dev-server.js",
+    "dev": "NODE_ENV=test node ./webpack/dev-server.js",
```
kprist: This should change, and replace it with `NODE_ENV=development` and also fix the check in email.js. Typically `test` is for `npm test`, not development server. Another alternative would be to remove it all together, and just assume development is the default if NODE_ENV is unset

services/email.js
```diff
@@ -0,0 +1,72 @@
...
+            });
+        }
+        catch (e) {
+            console.log(e);
+            return callback({message: 'Wrong SMTP configuration', error: e});
+        }
+
+        connection.on('error', (err) => {
+            console.log('ERROR on SMTP Client:', err);
+            if (process.env.NODE_ENV === 'test')
```
kprist: Again, this should be replaced with `!(process.env.NODE_ENV) || process.env.NODE_ENV === 'development'`

services/email.js
```diff
@@ -0,0 +1,72 @@
...
+            return callback({message: 'Failed creating connection to SMTP server', error: err});
+        });
+
+        connection.connect((result) => {
+            //Result of connected event
+            console.log('Connection established with result', result, 'and connection details (options, secureConnection, alreadySecured, authenticated)', connection.options, connection.secureConnection, connection.alreadySecured, connection.authenticated);
+
+            connection.send({
+                from: params.from,
+                to: params.to
```
kprist: We shouldn't expose the `to` parameter to the fetchr API, it should be a config parameter instead. If we need to support variant email recipients, we should provide a config parameter for each. Otherwise the service could be exploited for spamming anyone.

services/email.js
```diff
@@ -0,0 +1,72 @@
...
+            },
+            'From: <' + params.from + '>\r\n' + 'To: <' + params.to + '>\r\n' + 'Subject: ' + params.subject + '\r\nDate: ' + (new Date()).toGMTString() + '\r\n\r\n' + params.message,
```
kprist: It could be useful to also include here the user's name (if the contact form also collects that).

services/email.js
```diff
@@ -0,0 +1,72 @@
...
+            (err, info) => {
+                console.log('tried to send the email:', err, info);
+
+                try {
+                    connection.quit();
+                }
+                catch (e) {
+                    console.log({message: 'SMTP connection quit failed', error: e});
```
kprist: maybe this should be a `warn` instead of `log` ?

services/email.js
TBoonX: Is included in the message

TBoonX:
Please review again

services/email.js
```diff
@@ -9,9 +9,11 @@ export default {
     name: 'email',
     // At least one of the CRUD methods is Required
     create: (req, resource, params, body, config, callback) => {
-        if (!params.from || !params.to || !params.subject || !params.message)
+        if (!params.subject || !params.message)
             return callback('Error: Missing parameters', null);
 
+        const to = SMTP.to, from = SMTP.from;
```
kprist: I think you misunderstood me, or maybe I'm missing something here? It seems the person's email address gets lost along the way, so we have no way to know who they are or to contact them back ? I only said the `to` address should be a config parameter.
Also, I think we should avoid to simple names for local variables, like `to` or `from`, a better would toEmail, or toAddress and such ...

services/email.js
```diff
@@ -26,34 +28,34 @@ export default {
...
+            // console.log('Connection established with result', result, 'and connection details (options, secureConnection, alreadySecured, authenticated)', connection.options, connection.secureConnection, connection.alreadySecured, connection.authenticated);
 
             connection.send({
-                from: params.from,
-                to: params.to
+                from: from,
+                to: to
             },
-            'From: <' + params.from + '>\r\n' + 'To: <' + params.to + '>\r\n' + 'Subject: ' + params.subject + '\r\nDate: ' + (new Date()).toGMTString() + '\r\n\r\n' + params.message,
+            'From: <' + from + '>\r\n' + 'To: <' + to + '>\r\n' + 'Subject: ' + params.subject + '\r\nDate: ' + (new Date()).toGMTString() + '\r\n\r\n' + params.message,
```
kprist: I only suggested the `from` part of the email composition be embellished with the contact requester name, if available in the API. For example it could be: `'From: ' + params.fullName + ' <' + params.from + '> ' + ... ` ( if I remember correctly the email header syntax )

services/email.js
TBoonX: My thinking:
_From_ and _To_ have to be given by the config: _To_ relates to the email address where we want that each _contact us_ email should go. _From_ is the email address which is used as sender which just could be something@slidewiki.org, because imitating other email addresses will result in marking our domain globally as SPAM.
The action/component have to make sure that the message of the email contains everything. This is already done by Sole in the Contact Us component.
Using SMTP _To_ and _From_ is well defined but yes I could use longer names.

services/email.js
TBoonX: Please have a look at https://en.wikipedia.org/wiki/Simple_Mail_Transfer_Protocol
We have to follow the protocol regarding the first Headers. Whats inside the message is not restricted. Thats why Sole creates a detailed message: https://github.com/slidewiki/slidewiki-platform/blob/SWIK-1542-ContactForm/components/Home/ContactUs.js#L299

services/email.js
kprist: I see now about the from thing, OK, although I am not sure about the spam thing you mention, but I agree let's play it safe for now :)

services/email.js
kprist: I only suggested that because I had something else in mind, you are right of course. However in the wiki you can see
```
C: From: "Bob Example" <bob@example.com>
C: To: Alice Example <alice@example.com>
```
That's what I meant, and I thought the library supports that type of thing. But of course since both from,to are config params, no need for this either :)

PR #532, opened 2017-08-07
Swik 1542 contact form

TBoonX:
New route and component for Contact us
Calls the email platform service to send an email


Sole-Valero:
- Type of report is not required now.

- dropdown has been changed by and input element + datalist -> so readers are capable of linking the label and the input element

- datalist is marked with the role=listbox

- each option marked with role=option and aria-selected = true or false

- added code to control when an option has been selected

abijames:
@kadevgraaf @kprist I am ok for the to be merged from the UI side but I have not been able to test if the form submission works as the captcha won't work till on experimental. So if you can merge it we will retest.

PR #535, opened 2017-08-08
Swik 1287 translation ui

DaryaTarasowa:
This is content translation component
it will not work until:
1. Deck-service_1319 is merged to the master (@kostis)
2. translation service is deployed (@Paul)
3. Deck service AND the platform are configured to work with the translation service (@ben)
4. script from migration/fill_translations is run on the database (@ben)


kprist:
Of the steps Darya outlined:

1. pull request for translation support has been merged ( https://github.com/slidewiki/deck-service/pull/106 )
2. translation service has already been deployed and running

what remains is:
3. Deck service AND the platform are configured to work with the translation service (@ben) : not entirely sure what that entails
4. script from migration/fill_translations is run on the database (@ben): this one does not prevent the review for the commit, as long as you only try translating decks created from now on on experimental. this script applies to older decks, where some migration needs to happen in order to properly track all translations of a deck

kprist:
This means reviewing of the current PR is now possible if one checks out the branch in a local platform dev environment and point services to experimental. Remember to update the local `configs/microservices.js` with the translation service URL as is in `configs.microservices.sample.js`.

kadevgraaf:
@kprist @paulbaptista 
Darya took a look at this, but it does not work anymore.

kprist:
@kadevgraaf could you provide a bit more info as to what went wrong ?

kadevgraaf:
@DaryaTarasowa  - do we need to work out whether certain HTML brakes the translation (see my comment above) ? Or shall we already merge this and gradually improve it?

kadevgraaf:
@DaryaTarasowa  - one small thing - the translation button was planned to be above the decktree, next to the fork button. Can you add this here?

DaryaTarasowa:
@kadevgraaf yes, I'm planning to do that - yesterday I was working also on the second issue - the html that gets translated.

kadevgraaf:
Cool, many thanks!

kadevgraaf:
Works very well for small decks!
E.g., http://platform.experimental.slidewiki.org/deck/3000
http://platform.experimental.slidewiki.org/deck/3001
http://platform.experimental.slidewiki.org/deck/3002
http://platform.experimental.slidewiki.org/deck/3003
http://platform.experimental.slidewiki.org/deck/2994
http://platform.experimental.slidewiki.org/deck/3004

But gives 500/503 timeout error for for larger decks http://platform.experimental.slidewiki.org/deck/2992 
I think we might need to use cronjobs or some queue  + an email when translation was finished... @bwulff @rmeissn  @TBoonX - do you have any ideas about good options for this?

rmeissn:
@kadevgraaf You noticed that the linked decks are all a bit different, right?

You could do a request to translate deck XYZ and notify people via their notification area as soon as it finished.
It might be also good to sent an email to people about unread notifications once to twice a week (like facebook does).

kadevgraaf:
Hi @rmeissn 
Yes, they are results of translations.

Thanks for the ideas. That would be good indeed.
Currently we think the translation service crashes with large decks either
- because there is an internal server error 500 because translating specific slides results in an error.
- because it sends out too many requests (in this case a queue would be good)

kadevgraaf:
I have added a note that this is experimental, and want to merge it into experimental so that we can test it  and show it to trial partners

PR #536, opened 2017-08-09
Swik 235 236 show number of shares and downloads

dpaun:
Two more labels have been added to the DeckViewPanel, displaying number of shares and number of downloads of a deck. These numbers are retrieved from the activity-service.

### DIFF
--- a/actions/activityfeed/incrementDeckViewCounter.js
+++ b/actions/activityfeed/incrementDeckViewCounter.js
@@ -1,0 +1,8 @@
+import log from '../log/clog';
+
+export default function incrementDeckViewCounter(context,payload,done){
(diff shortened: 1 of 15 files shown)


abijames:
@dpaun could you make one addition improvement to this work prior to merging? Could you make the column displaying the labels in the deckview right aligned for better layout? This should be possible by changing line 108 of components/Deck/ContentPanel/DeckModes/DeckViewPanel/DeckViewPanel.js to <div class="right aligned column">

dpaun:
Yes, of course. Thanks for the suggestion.

abijames:
@kprist I am happy for this to be merged

PR #537, opened 2017-08-10
Fix partially hidden tooltip problem in decktree

stavmars:
The buttons in the deck tree used the css semantic ui tooltip which
in many cases (e.g. first node) wouldn't show correctly. This commit
rewrites the buttons using the corresponding react semantic ui components
and uses the react semantic popup for the tooltip, which resolves the
issue.


abijames:
@kprist could you merge this?

PR #538, opened 2017-08-10
Swik 1581 encode login  intl

TBoonX:
Exchanged texts in LoginModal and the corresponding actions with Intl ones


kadevgraaf:
Works well on localhost. Thanks!
@DaryaTarasowa  - should we add these int variables (e.g., payload.errorMessages.error403 , userSignIn.errormessage.isSPAM) to /intl/NN.json or in Transifex?

DaryaTarasowa:
@kadevgraaf I have added the new lines to transifex) I would prefer to check by myself the lines before adding. As it is not just adding - it is updating the file for translation (transifex merges the old and the new file automatically).

kprist:
@DaryaTarasowa could you provide some documentation for this process you mentioned? I think you also use a tool to automatically create the en.json file ?

PR #540, opened 2017-08-17
Swik 1601 - when forking a deck, show a link to the target deck in the activity feed

dpaun:
Additional information (target deck id) was added to the 'fork' activity which is created when a deck is forked. This was done in order to show a link to the new, forked deck, in the Activity feed.


kadevgraaf:
Works well, thanks!

PR #542, opened 2017-08-24
Swik 1085 new activities for using slides

dpaun:
New activities are created (using the activities-service) when slides/subdecks are attached (used).
Activity feed of the attached subdeck/slide displays afterwards something similar to: "dpaun used this slide in deck New deck"


kprist:
I saw this error on the platform when attaching a slide:
```
body: '{"statusCode":400,"error":"Bad Request","message":"\\"value\\" at position 0 fails because [child \\"use_info\\" fails because [child \\"target_name\\" fails because [\\"target_name\\" is not allowed to be empty]]]","validation":{"source":"payload","keys":["0.use_info.target_name"]}}' } }
```
But only twice, tried to recreate it with the same slide but couldn't, not sure what happened

dpaun:
I've made some small corrections to this branch and the activities-service, so that this error does not happen again.

PR #544, opened 2017-08-25
Swik 1622 get datasources from already existing deck data - Performanse improvement

dpaun:
Data sources are already a part of deck/slide data, retrieved when getting the content. We will use this instead of calling the deck-service again to get data sources.
This should somewhat improve the platform performance.


kprist:
I disagree with this approach in general. Sources are the default tab in the deck page by coincidence, so if we change the UI this improvement will be lost. Furthermore, I would like to see more data as to how much of an improvement this will prove to be, could we have those ?

I would be interested to see what would happen if actually removed the datasources information from the deck call, in general make the fetch deck response leaner. It would benefit more I believe, as it is used in other circumstances. Also, since react components are rendered somewhat independently, removing the datasource info from the deck info call might improve performance more than this approach does. 

I would recommend to hold this PR until we've seen numbers and also tested the alternate approach. I would also like @ali1k to comment on this as well.

dpaun:
I don't have any measurements yet.
I agree that we should consider whether it would be better to remove the datasources information from the deck call and have a separate route for getting the datasource list (and count).

kprist:
I've started some exploratory work on branches https://github.com/slidewiki/deck-service/tree/SWIK-1627_Refactor_datasources_api and https://github.com/slidewiki/slidewiki-platform/tree/SWIK-1627_Refactor_datasources_api . I can see some code from this branch that could be combined with those branches so please keep this PR open until I've completed that.

dpaun:
Closing the PR, since it's been replaced with https://github.com/slidewiki/slidewiki-platform/pull/558 .

PR #545, opened 2017-08-25
Swik 1618 do not unnecessary load notifications list

dpaun:
Instead of loading the complete list of new notifications, only their count will be fetched in order to inform users if there are any new notifications.
Also, query parameters are introduced in calls to notification-service and discussion-service, utilizing the latest changes in their API.


kprist:
Numbers are good :) Loading homepage included a call to fetch notifications about 3sec, now the count is down to 600 ms ( still a bit slow but that can be improved on the backend, or maybe experimental is slow in general ). :+1:

PR #550, opened 2017-08-30
SWIK-1630 improve server side rendering performance

rmeissn:
According to https://www.youtube.com/watch\?v\=PnpfGy7q96U , these changes should improve the server side rendering performance a little bit. I wasn't able to test the improvement, as we have no profiler/reliable performance test and a manual test has large deltas.

Please Note: webpack and babel are required as of the production deployment and are thus located wrong in the "devDependencies" section of package.js. I haven't changed this!


rmeissn:
One questions: Do we need all of these babel plugins? Because they haven't been used until now...

rmeissn:
package.json dependency considerations have been resolved by another PR and have been merged into this branch.

PR #553, opened 2017-09-04
Swik-1636 cookie on all domains bug

TBoonX:
I removed all domain entries inside the userStoragePlugin
Testing is quite hard. I did test it with localhost, thus cookies are working in general but we will see If the issue is fixed


kprist:
@TBoonX if you want more people to test this, could you list a few steps we could take ?

TBoonX:
To test it properly, you need a domain and a subdomain (where you can run the code) to test the cookie on. Signing in in the domain should create a cookie with the domain as domain value. If there is a leading dot then this fix does not work. After creating the cookie the subdomain should be visited. If there is also a cookie on the subdomain active this fix didnt help. And vice versa.
In the unfixed version of the code the domain value was .domain    -> the leading dot makes the cookie valid for the domain and all its subdomains.

kprist:
@TBoonX maybe then we can test this by putting some domains like you described in my `/etc/hosts` file ?

TBoonX:
@kprist Should work

kadevgraaf:
Thanks!
Looks good when logging in/out between experimental and localhost. 
Can we further test this between experimental.slidewiki.org and testing.slidewiki.org?

PR #556, opened 2017-09-06
Avoid loading all slides in deck route

stavmars:
When navigating to a deck/ route in the platform, it would fetch all slides (metadata + content) just to pick the `theme` attribute for the deck tree. This pr changes this and uses another store for getting the theme value.


kadevgraaf:
Works well. Thanks!

PR #558, opened 2017-09-11
Swik 1627 refactor datasources api

kprist:
This (if approved) is meant to replace https://github.com/slidewiki/slidewiki-platform/pull/544.

It includes work from that pull request, so that less requests are made, but also depends on the merging of https://github.com/slidewiki/deck-service/pull/108, where the GET /deck/:id API does not return the data sources.


kprist:
In order to test/review this branch, one needs to also use the branch from https://github.com/slidewiki/deck-service/pull/108 . The deck service branch once merged will break slidewiki platform, so these two PRs need to be merged at the same time.

kadevgraaf:
Thanks! Merged together with https://github.com/slidewiki/deck-service/pull/108

PR #517, opened 2017-07-17
Swik 1451 create functional interface for edit question

vinaymodi:
Edit question works fine. Question that appears in the question panel can be edited by the deck's owner. Question edit interface allows editing all fields of a question: title, difficulty level, answers, correct/incorrect, explanation. When save button is clicked, this information is sent to the service on the platform.

At the moment, PUT api in questionservice microservice is not working (I get error 503) therefore one can't see the question getting updated into database. Once that service is fixed by respective developer, you should be able to see question updated in the database.

Branch 1427 is also merged into this work. That (swik-1427) branch is a dependency for this work. PR for swik-1427 is already raised.

PR #562, opened 2017-09-14
Swik 1608 delete a question

vinaymodi:
Kindly test the branch. I have added delete question functionality. Also made minor updates into add/edit functionality. Earlier after adding/editing a question, the add/edit interface was shown as it is. But now, it will again show the updated question panel (i.e. after add/edit done).

A proper testing of this would need delete API to work. But without it also it should be possible to see Delete button and dialog box etc.


kadevgraaf:
Thanks!
I will merge 1451 - https://github.com/slidewiki/slidewiki-platform/pull/517 first and then merge this into master after that

dpaun:
I propose that we close this PR, since deleting of questions is also implemented in Swik-1602 branch (the main branch containing questions functionality implementation) which also has a PR open.
Looking at Files changed, there doesn't seem to be anything which does not exist in Swik-1602 branch.
@kadevgraaf , do you agree?

kadevgraaf:
Yes I agree. Sorry that I could not do more on this and thanks for the hard work @dpaun !

PR #564, opened 2017-09-18
    SWIK 1705      Change emitChange() in slide content editor to check if original html has changed to current html

kadevgraaf:


kadevgraaf:
Already fixed in newer PR

PR #565, opened 2017-09-18
seems to be parallel

DaryaTarasowa:
I've checked the major routes and the related to them actions. If I understand correctly, async.parallel will run the actions in parallel. By the benchmarkings I have noticed neither significant improvements, nor (luckily) decrease of performance.


configs/routes.js
```diff
@@ -247,7 +247,7 @@ export default {
         page: 'deck',
         handler: require('../components/Deck/Deck'),
         action: (context, payload, done) => {
-            async.series([
+            async.parallel([
```
kadevgraaf: @kprist @mmeimaris - is this a good change? Or is there a good reason it should be series instead of parallel?

kadevgraaf:
Looks good. I approve.

configs/routes.js
kprist: I think @stavmars or @huwf might have a better idea about this, they are updating stuff over in other branches that may make this PR obsolete

stavmars: @kadevgraaf this https://github.com/slidewiki/slidewiki-platform/pull/556 actually makes your change unnecessary.

kprist:
Relevant code removed with https://github.com/slidewiki/slidewiki-platform/pull/556

PR #567, opened 2017-09-18
Translation in Transifex: architecture changed to default.json

DaryaTarasowa:


kadevgraaf:
Looking good on localhost. Thanks!

PR #568, opened 2017-09-19
Swik 1584 subdecks only stripped down

huwf:
If the user is viewing a subdeck, or a deck within a subdeck, only these slides will be loaded into presentation mode.  Requires #556


huwf:
And I also added a string replace in the presentation component, which changes the content to be  instead of .

kadevgraaf:
looks good! Many thanks!

PR #570, opened 2017-09-20
Fix change log breaking when deck metadata updates are included

kprist:


kadevgraaf:
Thanks for the fix.
I get a 400 on localhost (deck/2074/deck/2074/ on deckservice.experimental) when changing deck metadata (tag/description field/title) of a deck. Or should I test in some other way? 
screen shot 2017-09-25 at 17 52 52

kadevgraaf:
Thanks! Works like a charm now.

PR #571, opened 2017-09-21
SWIK-1487-shorter-descriptions

rmeissn:
Restricted quantity of allowed chars

Is this a good solution?

For tests: visit http://localhost:3000/user/72


abijames:

PR #572, opened 2017-09-21
Fix sending filters after clear in search form

schatzopoulos:
Fixes bug that appears when clearing a search filter


abijames:
bug is fixed @kadevgraaf this can be merged

PR #573, opened 2017-09-21
Swik 1643 update platform with deck and import secured apis

kprist:
This depends on two other PRs, 
https://github.com/slidewiki/deck-service/pull/110
and
https://github.com/slidewiki/import-service/pull/32

All should be tested locally and merged together once approved.


kadevgraaf:
Thanks!

PR #575, opened 2017-09-24
SWIK 1511 1512 1513 1514 problems with identifiying ID of input elements

kadevgraaf:


kadevgraaf:
@abijames  - can you check this commit. There is a lot of fixes and (hopefully) improvements in here, also for accessibility/keyboard shortcuts.
@bwulff and @paulbaptista   - can we have this in the release?

kadevgraaf:
Hmm, I discovered some minor things:
In general: When you click on an input box and move it with arrows (ctr+alt+up/down/left/right) sometimes an earlier input box is moved
In firefox and safari: the drag and context menu buttons are not placed at the top-left - can fix this quickly
in firefox and safari: duplicate does not always work well - sometimes only the menus are duplicated
In firefox: selecting text with mouse does not always work (caret is reset)

kadevgraaf:
Fixed:  In firefox and safari: the drag and context menu buttons are not placed at the top-left - can fix this quickly 

Not fixed: In firefox: selecting text with mouse does not always work (caret is reset) - this is a larger/more difficult issue with the combination CKeditor and Firefox - requires some hacking

kadevgraaf:
Fixed duplication problems

PR #576, opened 2017-09-25
Bugfix SWK-1678 load on selected slide

huwf:
This adds the location.hash to the page in the Presentation component, which allows reveal.js to parse the URL to load on the specific slide.  Requires #568


huwf:
The branch name refers to the wrong JIRA ticket, but I have modified the title to be the correct one.

kadevgraaf:
Looks good. Thanks! Tested on localhost. See e.g. http://localhost:3000/deck/2298-1/slide/13867-1/13867-1:28/edit

PR #582, opened 2017-09-28
Swik 1704 using activity service query params

dpaun:
The latest changes to the activities-service introduced query params (backward compatible code) in order to simplify the API.
A few small changes to the platform in this PR modify fetching of deck/slide activities (for activity feed) and share/download/like counts, by making the calls utilize the new activity-service API.


kadevgraaf:
Works well on localhost. Thanks!

PR #584, opened 2017-09-29
Swik 1716 use react-share for email sharing

dpaun:
The new version of react-share package supports sharing by e-mail, which was previously implemented as a separate component on the platform.


kadevgraaf:
Works well! 
A minor comment, that is related, but not about this actual code, is that the activity says "<username> shared this slide", and then in the overview of activities on deck level it also says "<username> shared this slide". I have made a SWAQ ticket for this.

PR #585, opened 2017-09-29
Refactor code to include webpack build phase into docker image build

kprist:
With this effort the goal is to improve the time it takes for a docker image with the platform to run. Currently, when starting the container, we issue a full webpack build that bundles the webapp assets. The main issues we have with this, is that it takes some time for the container to start (or more importantly, to restart). Also, if there's an error in the webpack build phase, it only shows when starting the container, whereas it should happen sooner, during the build phase.

The reason we included the webpack build phase in runtime is because we need to respect environment variables set when starting the container, and those ENV variables changed some files that are part in the webpack bundle.

The solution presented here splits the webpack bundle to two phases: one with all the code that does not depend on env variables, and one with those files that need to be modified according to the environment. The latter produces a helper library bundle that the former simply references. The way they are set-up is a bit unorthodox, but the aim was to edit as few files as possible, and also to not interfere in any way with the dev process of the platform, i.e. `npm run dev` is virtually untouched as a process.

This has been tested locally and should work for deployments to Docker hub as-is, but needs to be tested. Also, in theory it should work on travis as well, but that should also be verified.


rmeissn:
Good Idea to split the build process into two phases.
Both phases should be added (or at least the first one) to the Travis tests (script), as the process is mandatory to execute the software. Latter one won't work, if the build errors and should be reported via mail/slack too.
I am not really familiar with webpack build processes (I think Ali has better insights/comments). Code quality is very good. 
Why do you need to add the env vars directly to the Dockerfile? If the first build phase won't need any env vars, I see no reason to add it?!
Haven't tested it yet by executing it.

kprist:
If you need to change anything in travis, please add the lines that should be run, I haven't touched it as it seems all it did was run linting and tests. Docker build happens only when updating `master` if I'm not mistaken, so if the build errored before this change, how did it report that? Or did the build never threw an error before?

I didn't want to change much of what `npm run build` does, so even though i split the webpack process in two, the `webpack/prod.config.js` file includes essentially both processes, i.e. the bigger build and the env-dependent one. The latter uses the config in `webpack/prod.settings.config.js`, which is imported in `webpack/prod.config.js`.

I added the env vars in the Dockerfile essentially because I wanted a way to have `docker run .` working out of the box. I considered `docker-compose up` where the default env vars are included in `docker-compose.yml` but I am not sure how I would go about running it locally, as the yml file references the dockerhub image, i.e. I would need to push my local build first (?). I just want to run it locally using docker and pointing to the services on experimental, no docker hub etc, is that possible without putting the env vars in Dockerfile ?

rmeissn:
@bwulff The changes in the Dockerfile will introduce default URLs for all services, default secrets and email settings, ... for the docker image. You would end up with "Override everything via env vars as of docker run (or compose). If you forget to override these, the container will use the default settings". The last time we introduced default settings, you raised your hands against it. Do you still insist on it?

@kprist npm run build phase separation makes sense. About Travis: Yes, it only executes one test and the linting process. It errored before for e.g. linting errors and reported these via mail and in Trello (frontend channel). I like to add the `npm run build:nostart` to the Travis tests, as this must finish successfully, otherwise the image is broken and shouldn't be build and uploaded at all. Thus I think it's a good addition to the Travis tests.

rmeissn:
Will fix the Travis issue tomorrow...**grrrrr**

bwulff:
Confirmed that branch `improve_docker_startup_time` builds successfully on Docker Hub
https://hub.docker.com/r/slidewiki/platform/builds/bltyx48qglrknmni3fewpnb/

bwulff:
Works on testing:
https://testing.slidewiki.org/

Starts considerably faster!

bwulff:
@rmeissn @kprist : Roy is right, hard-coded default URLs were a constant pain in the butt --> because when wrongly configured the software quietly does something unforeseeable e.g. connecting to services / the database on experimental. And we don't want that.

However I understand there is a docker-compose.yml for each service that is used to deploy the Travis build to experimental
https://github.com/slidewiki/slidewiki-platform/blob/improve_docker_startup_time/docker-compose.yml

Could we just leave the ENVs in the Dockerfile at a default value like "SET_ME", since the values for the ENVs should be provided by the compose file?

rmeissn:
@kprist The way I've added the implemented process to Travis doesn't requires any env vars (except for a SMTP port), as most files are already valid json files. Thus you wouldn't introduce any default URLs/vars.

kprist:
@bwulff @rmeissn I've removed the experimental default URLs and just left the variables listed but empty; the build is successful and I've also tested that `docker-compose up` properly overrides those empty values.

ali1k:
I am not much into the docker deployment procedure we have. Nonetheless, from my understanding, we are unnecessarily running the heavy Webpack build process each time the docker container starts, only because of the run-time env variables that we need to apply. If it is the case, it absolutely makes sense to separate those configs from our main bundle and add them at run time. The Webpack config that used externals looks good. I just fixed some template issues for variables and resolved minor lint issues. Also synced it with master.
After this change is merged, I am gonna upgrade to Webpack 4 which might also affect the performance for bundling...

@rmeissn @kprist @bwulff I think this PR is ready to merge. I cannot test the docker build process, if you think it is fine, please go ahead and merge it to master asap as it affects the performance a lot.

rmeissn:
I've tried to built it and it errors with:

```
npm WARN lifecycle husky@0.14.3~uninstall: cannot run in wd %s %s (wd=%s) husky@0.14.3 node ./bin/uninstall.js node_modules/husky
npm WARN ajv-keywords@3.1.0 requires a peer of ajv@^6.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN extract-text-webpack-plugin@2.1.2 requires a peer of webpack@^2.2.0 but none is installed. You must install peer dependencies yourself.
npm WARN postcss-modules-resolve-imports@1.3.0 requires a peer of postcss@^6.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN The package css-loader is included as both a dev and production dependency.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

...

(node:18) DeprecationWarning: Chunk.modules is deprecated. Use Chunk.getNumberOfModules/mapModules/forEachModule/containsModule instead.

...

Child settings:
    Hash: b868212c21f0c0d1b837
    Time: 1807ms
          Asset     Size  Chunks             Chunk Names
    settings.js  1.81 kB       0  [emitted]  main
       [0] ./configs/settings.js 980 bytes {0} [built]
       [1] ./configs/microservices.js 1.19 kB {0} [built]
       [2] ./configs/secrets.js 321 bytes {0} [built]
       [3] ./configs/general.js 166 bytes {0} [built] [failed] [1 error]
    
    ERROR in ./configs/general.js
    Module build failed: SyntaxError: Unexpected token (9:0)
    
       7 |     loglevel: 'debug',
       8 |     ssoEnabled: 
    >  9 | };
         | ^
      10 | 
    
     @ ./configs/settings.js 17:15-35
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! slidewiki-platform@2.0.0 build:nostart: `webpack --config ./webpack/prod.config.js`
npm ERR! Exit status 2
npm ERR! 
npm ERR! Failed at the slidewiki-platform@2.0.0 build:nostart script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2018-04-23T08_43_06_625Z-debug.log
```

kprist:
@rmeissn since it worked on travis, I'm guessing you tried simply building the image (without docker-compose), so the SSO_ENABLED was missing etc. I added a default env for that ( I think it's correct this way ).

There was another issue with that however in the travis build with general.js where envsubst did not work correctly so it fails on travis. @ali1k tried to fix it but created another potential issue with the use of SSO_ENABLED flag. So I think the issue is with the `travis.yml` file and what it does in "before_script". We need to have proper envsubst in there somewhere I think, but I'm not sure what your intention writing that was, as you only seem to care about doing envsubst with the secrets.js template, not the other ones. Could you check that out ?

rmeissn:
@kprist You're probably right. Forgot to execute something...
I only cared about the specific problem that occurred. If it's more clean to do this for all these files, feel free to change the mentioned line.

rmeissn:
I think this is fine to merge - let's have the changes in master to not fix this branch again and again. Are you also fine with this, @kprist ? If yes, please merge.

PR #588, opened 2017-10-03
BugFix: Swik 1599 fix download with a dialog

Sole-Valero:
Download button opens a dialog in which users can select the file format to download, and then the user can click on "download"  or on "cancel" to discard the action.

With this new interface, the bug instroduced by the download button is erased.

It also uses intl to translate the sentences, except the name of the formats.


Sole-Valero:
Last commits solve the following issues:

- the cancel button shows "cancel" as default string
- the label which contains the description of the modal now is focusable, so when the modal is open the tab goes to it
- added a missing aria-checked label in one of the options in the radio group

abijames:
@kadevgraaf I have tested this already so ok to merge

PR #590, opened 2017-10-05
Swik 1365 fix issue of inconsistent focus in decktree

stavmars:
This PR makes some accessibility improvements in the deck tree such as using tab to tab into and out of the deck tree and arrow keys to change the focused tree node.


abijames:
@kadevgraaf  I have tested this and ok to merge

PR #591, opened 2017-10-05
added resolve/alias in web pack config

kadevgraaf:
there is a small chance of that the libraries that depend on an older react version (which causes the IE problem, and is resolved by this PR which has a Webpack resolve of react library versions) might have different behaviour with a newer version of React…
I already checked critical functions/features quickly on my localhost (could not duplicate/attach decks/slides because of timeouts, but this might not be due to this change), but to be absolutely sure we should again test/check the critical functions/features after this fix…


abijames:
Error messages on testing in IE 11:
Home page:
Line: 52
Error: Minified React error #39; visit http://facebook.github.io/react/docs/error-decoder.html?invariant=39&args[]=%20This%20may%20be%20caused%20by%20unintentionally%20loading%20two%20independent%20copies%20of%20React. for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
 
 
 
Deck listing:
Line: 52
Error: Minified React error #39; visit http://facebook.github.io/react/docs/error-decoder.html?invariant=39&args[]=%20This%20may%20be%20caused%20by%20unintentionally%20loading%20two%20independent%20copies%20of%20React. for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
 
Deck page:
 
Line: 52
Error: Minified React error #39; visit http://facebook.github.io/react/docs/error-decoder.html?invariant=39&args[]=%20This%20may%20be%20caused%20by%20unintentionally%20loading%20two%20independent%20copies%20of%20React. for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

kadevgraaf:
on http://facebook.github.io/react/docs/error-decoder.html?invariant=39&args%5B%5D=%20This%20may%20be%20caused%20by%20unintentionally%20loading%20two%20independent%20copies%20of%20React it says:
ReactDOM.render(): Invalid component element. This may be caused by unintentionally loading two independent copies of React
So still the original error

kadevgraaf:
I will merge this. It seems like a good idea to do anyway, and it might a prerequisite for a more elaborate fix.

PR #592, opened 2017-10-05
Swik 1580 encode header for translation

DaryaTarasowa:
Here are new lines from transifex (including proper English ones). Also, I have separated the intl stuff from production run. Do not need it in there) But I had to separate also the .babelrc plugins development from production - because in the production we do not need intl plugin, which conflicts with "transform-react-inline-elements"


kadevgraaf:
Looks good on localhost with both dev and production mode.

abijames:
This PR has changed the layout of the user menu so that is horizontal not vertical. When you close the menu there is also a diamond shape left on the screen indicating the menu is not rendering properly.
@Sole-Valero  and I also trying to fix accessibility on this menu so we are going to get conflicts between these changes https://slidewiki.atlassian.net/browse/SWIK-1500

screen shot 2017-10-09 at 13 59 39

PR #593, opened 2017-10-06
SWIK-1732-fix-local-docker-build

rmeissn:
* Reverted PR #579
* Fixed some warnings of npm (missing packages, wrong versions)
* Added fixes from microservice-template [PR#4](https://github.com/slidewiki/microservice-template/pull/4)

Docker should now be able to build locally without any issues, even though Travis is building docker images much faster. Please review carefully for the existing workflows.


kprist:
@rmeissn could we please review this in conjuction with my work on https://github.com/slidewiki/slidewiki-platform/pull/585 ? I actually wanted to suggest you take a look at my proposal and apply the travis speed-up in my branch, but you got me to it. I did add you as a reviewer there, so I would appreciate your input.

rmeissn:
I've read through PR #585 . The PR will not interfere with this one and won't benefit from using Docker build args (I think). Do you see any other imbrication's?

kprist:
No I don't actually, I just wanted to make sure you didn't either.

.dockerignore
```diff
@@ -1,4 +1,3 @@
-node_modules
```
bwulff: why don't we want to ignore the node_modules dir anymore?

rmeissn: Because: 
1) Travis checks out the repo (no node_modules installed)
2) Travis installs/updates all npm dependencies (node_modules available + up to date)
3) We are ignoring node_modules for the docker build
4) We're installing/updating all npm deps as of the docker build

Last one can be skipped if node_modules are not ignored anymore and thus speed up the docker build step by ~1-2mins.

kprist: To be more precise, docker build does not ignore the `node_modules` directory, but it removes it before build when not in travis context. This means that when you build the image outside travis, if you have any node modules installed in the directory, they will be copied over in the build context, and then removed. Maybe this could be improved by changing the line `ADD . ./nodeApp` to not copy the `node_modules` directory? Not sure if you can, or if it improves anything really, it's a small thing anyway.

Speeding up the build process by not ignoring the node_modules may be worthwhile, but I don't think it applies to the docker hub context. Also, ignoring/removing the node_modules will result to a more predictable image building, especially after we upgrade to npm 5 and start using locked package versions.

rmeissn: I don't see why removing the node_modules folder will result in a more predictable image....At least for Travis, we're installing always the latest available version of a package. If you built it locally, it does the exact same.
Btw.: You can also restrict package versions now as of the package.json file, by not using the '^' operator or exchanging it with another operator (like `~`). The current default behaviour (`^X.Y.Z`) is to update to the latest minor version, but not update any major version (v major.minor.patch).

Removing node_modules folder or not including it at all will only result in 1-4 secs. I think it's not worth the effort.

rmeissn:
Ready to merge?

kadevgraaf:
Works for me. Many thanks!

PR #594, opened 2017-10-10
Swik 1756 Fixing migrated datasources

dpaun:
This is a very small, but very important fix for a certain problem with migrated decks.
Datasources from migrated decks (slides) contain _id attributes, which cause validation errors when saving slides.
This fix removes these attributes from datasources.


kprist:
It seems the platform simply reads whatever the service returns, keeps it in memory, then after a possible edit re-sends it to the deck service and that's where the problem occurs? Since it's the deck service that complains about it invalid data the service itself returned, I'd rather we fixed this on the deck-service itself and keep the platform code clean.

dpaun:
Yes, that is how the problem occurs. I agree it makes sense to fix it on the deck-service.

kprist:
Closing this at the issue will be fixed on service: https://github.com/slidewiki/deck-service/pull/115

PR #595, opened 2017-10-10
Swik 1472 do not add wrong items to the activity feed

dpaun:
This is a small bug fix to prevent wrong items in the activity feed to temporarily appear (e.g. 'like' activity belonging to the parent deck of a selected slide).


kadevgraaf:
Looks good. Thanks!

PR #596, opened 2017-10-10
Swik 1753 fix notifications loading message

dpaun:
'Loading' message for notifications was re-implemented to match how it's been done in some other components.


kadevgraaf:
On my localhost, the notifications keep loading, regardless of whether I have 25 notifications or none. This is however the same on https://platform.experimental.slidewiki.org/notifications . Maybe this has to do with some service being down?

screen shot 2017-10-18 at 12 45 00
screen shot 2017-10-18 at 12 44 38

kadevgraaf:
Works well. Thanks!

PR #600, opened 2017-10-12
Fix user not included in delete activity

stavmars:


kadevgraaf:
Indeed a nice and short fix :)

PR #601, opened 2017-10-13
SWIK-1763-fixContactForm 

Sole-Valero:
updated the package json to add the latest verion of semantic-ui-react:0.74.2
In previous version input lists are not well rendered, dropbox lists options are not read aloud, and more problems. 
Now, the input list shows the possibilities to the user in the ContactForm. You can check it in http://svalero.dsic.upv.es:3000/contactus
In the original branch in which the contactForm was implemented, also this entry was updated. At a later point, this change was reverted.


abijames:
@kadevgraaf this can be merged. i have tested

PR #603, opened 2017-10-17
Swik 1745 introduce attach activity type

dpaun:
When a deck/slide is attached, an 'attach' activity is created, instead of an 'add' activity.


kadevgraaf:
Works well. Thanks!

PR #605, opened 2017-10-20
Swik 1796 fix deck report

Sole-Valero:
- Now the deck report uses our mail service to send the report to the atalassian email address. 
- Focustrap updated: options were bad introduced
- Added more aria attributtes to the fields of the form
- The first focused element change to force readers to read aloud the modal description
- If the modal is closed with fields in error status, the status is reset to nomal status. So, when the user opens the dialog another time, it don't show any error (previously, one time the modal entered in error status, it didn't exit from that status) 
- Modal now is multilanguage (uses intl)

We need smpt service running  in the platform server in orther to test it well. I think it's only running  on stable or slidewiki.org


Sole-Valero:
@abijames , please can you add the informatin about your testing?

abijames:
@kadevgraaf I am happy for this to be merged. There are some accessibility issues on the dropdown list but we will deal with thees in a separate ticket.

PR #606, opened 2017-10-24
Support sending JWT token when posting user activities

kprist:
This completes work related to securing the activities- and xapi-services with jwt and including the email in the jwt token so that it can be safely and efficiently used in the xapi-service to properly write user activities in the learning locker.


kadevgraaf:
👍

PR #611, opened 2017-10-31
use dynamic domain based on microservice definition

kadevgraaf:
image upload expects that fileservice runs on same domain, otherwise
different, e.g., Cross-Origin Resource Sharing, method is necessary


TBoonX:
When the file-service is ready this works on other domains than slidewiki.org

PR #616, opened 2017-11-02
Loading indicator - via react states store and dispatch

kadevgraaf:


TBoonX:
Why do you request a review and then just merge it?

kadevgraaf:
Sorry. We looked at it and decided to merge to prevent conflicts with ongoing work.

PR #623, opened 2017-11-10
SWIK-1790 adding move activity

dpaun:
When a slide or a subdeck node is moved, a 'move' activity is created and shown in the activity feed.


kadevgraaf:
Works well! Thanks!

PR #629, opened 2017-11-15
Added a custom HTML template for the presentation rooms feature

rmeissn:
I've added a custom HTML template for the presentation rooms feature, just like we did for the presentation mode. I've replicated the introduced changes for the presentation mode and removed a lot of unnecessary JS libs and CSS libs, that are not needed for presentation rooms feature. 

This shrinks the size of the application (that initially must be downloaded) and the DOM size a lot. It also decreases initial load time of the page, as the browser does not need to evaluate so much libraries anymore.

Please verify that everything is still displayed correctly (as I removed some CSS libs) and verify that everything still works, without throwing any errors on the browser console (as I removed a lot of JS libs).


kadevgraaf:
Works nicely. Thanks!

PR #630, opened 2017-11-15
fix the bug related to toLowerCase

ali1k:


ali1k:
the bug stemmed from the translation panel where the value was not yet set and then toLowerCase was applied. I added some checks to only apply this function if the string is set.

kadevgraaf:
Works well, thanks!!

PR #642, opened 2017-11-15
Swik 1602 create a new question

kadevgraaf:


dpaun:
Latest commits have fixed existing and added missing functionality for question adding, editing, deleting.
There are still some things to be implemented, like question reordering, aggregation at deck level (if we want this), but I think this should go to some new Jira tasks.
@kadevgraaf @bwulff @abijames , I think this prototype can be merged.

kadevgraaf:
works well and looks good, thanks! I am happy to merge this.

PR #644, opened 2017-11-17
SWIK-1817 Upgrade to NodeJS v8

rmeissn:
I've upgraded the NodeJS runtime to v8 (currently LTS) on Travis and for the platform-runtime.

Works very well on my computer and seems to be much faster than with NodeJS v6.
Still testing, do not merge as of now.


rmeissn:
I've built the docker container locally and it seems to be executed as expected.
Seems to be good to merge.

Please test for yourself!

kadevgraaf:
Tested on localhost and seems to work well for the main/core functions.
I've created a task for the remaining packages -> https://slidewiki.atlassian.net/browse/SWIK-1942 and tagged this for the next release 2.8.x

PR #645, opened 2017-11-17
Fix bugs of presentations rooms

rmeissn:
Added a missing library - dropdown list of peers in "Ask a question" works now.

Maybe wait for more bugs to arise.

### DIFF
--- a/components/PresentationRoomsHTMLLayout.js
+++ b/components/PresentationRoomsHTMLLayout.js
@@ -27,16 +27,10 @@ class PresentationRoomsHTMLLayout extends React.Component {
             </head>
             <body>
                 <div id="app" aria-hidden="false" dangerouslySetInnerHTML={{__html: this.props.markup}}></div>
-                {/* Following are added only to support IE browser */}
-                <script src="/es5-shim/es5-shim.min.js"></script>
-                <script src="/es5-shim/es5-sham.min.js"></script>
-                <script src="/json3/lib/json3.min.js"></script>
-                <script src="/es6-shim/es6-shim.min.js"></script>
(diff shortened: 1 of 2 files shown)


kadevgraaf:
Hi @rmeissn 
Thanks for the fix!
Shall I/we create a testing task to identify possible bugs? 
Nevertheless I think it works already quite well - we demo-ed it during the plenary, and that went very well :) 
Maybe instead we can ask around for possible quick improvements? (then I think a different PR would be better to keep overview)

rmeissn:
Good suggestion to create a testing task and maybe have @abijames review texts (of modals) and usability aspects + provide improvement suggestions. I remember her mentioning that texts seem too technical.
It would be also good to have suggestions which elements to hide for mobile users - should be easy to implement as this is kind of a standalone feature. I was thinking of hiding the "chat" + remove some explaining texts for the transcript, in order to fit onto a 5" 16:9 screen without the need to scroll.
Asking for quick improvements is also a good idea. I already took some notes from the plenary about possible improvements.

rmeissn:
Ah, a question: There are [these lines](https://github.com/slidewiki/slidewiki-platform/blob/887b21611fde7b2b35dc77730f439e04eeead79d/components/PresentationRoomsHTMLLayout.js#L30) in the HTML template which are used to support IE11. Presentation Rooms won't run on IE11 anyway, as it is not supporting WebRTC. Is it safe to get rid of these lines?

kadevgraaf:
They should be safe to remove 👍  We decided during the plenary that we will not support IE11.

kadevgraaf:
Works well! Thanks!

PR #647, opened 2017-11-17
NPM Package Upgrades

rmeissn:
Ali and Roy upgraded all npm packages that were safe to upgrade and did not need "complex" code changes. We've created a document at Drive named "upgrade/test npm packages" that covers remaining packages and who we should contact to upgrade these.

Please test slidewiki-platform default use-cases (logged out and logged in) to verify that everything still works.


kadevgraaf:
Tested on localhost and seems to work well for the main/core functions.
I've created a task for the remaining packages -> https://slidewiki.atlassian.net/browse/SWIK-1942 and tagged this for the next release 2.8.x

PR #661, opened 2017-11-24
Swik 1684 delete notifications from list

dpaun:
In order to display notifications, the platform is currently getting new (unread) notifications from the notification service, while old (marked as read) notifications are retrieved from the activity service. This PR corrects this. The new version of the notifications service (already merged, backward compatible) does not delete notifications from the db, when users mark them as read, but only changes the value of a flag. This enables some improvements in the platform UI, like adding the possibility for users to delete notifications, which is also implemented in this branch.


kadevgraaf:
Looks very good! I am happy with merging this. @abijames is the usability/UX good?

dpaun:
The buttons are now improved, after the latest commit.

abijames:
@dpaun thanks for updating the buttons, they are now keyboard and screen accessible. 

I have a couple of further comments which don't need to stop this PR unless you can do them quickly:
* the information and buttons after the notifications title allowing you to mark all as read and delete all should by styled the same as the buttons on the notification (and will then be accessible). They could be placed on the line below the title.
* can you add tooltips to the buttons on the notification as you currently have on the mark all buttons?
* once you mark a notification as read or delete then focus is not managed. So when you delete a notification, focus goes on to the delete button for the next notification. It would be better for focus to be at the beginning of the next notification.

abijames:

PR #664, opened 2017-11-29
comment out translation panel, reinstate language label

abijames:
Please test! Comment out translation panel on deckview, added back in old code


kadevgraaf:
Works well. Thanks!

PR #666, opened 2017-11-29
Swik 1471 apply proper themes in slides and thumbnails

kprist:


Sole-Valero:
I show possible future problem with the changes made in the store used by AttachSubdeckModalStore.
In file stores/AttachSubdeckModalStore.js, please don't change the current logic of the property deckSlides , which should only contain the id of the slide. Create and use later a new property to store the themes. 
So, line 158 return(slide.id + '/' + slide.theme); --> come back to return(slide.id)
And add : 
this.deckSlidesThemes = payload.slides.map((slide) =>
{ - + return(slide.theme); }
);
Regarding slide.theme, I haven't found the modifications in the action to retrieve this data...where is it?

Sole-Valero:
Now, thumbnails are not diplayed in deck list, and slide grids.

Sole-Valero:
For attaching slides, the currect action which loads the data of the slides only retrieve the info about title, content, user, id an type of the slides. The field theme is not loaded.  In fact, the action loadSlides calls the service deck.slides. This service access to the deckservice.experimental.slidewiki.org/deck/{id}/slides, and only returns the values in the childern json property, which don't contains the theme.

Sole-Valero:
Also the webservice deckservice.experimental.slidewiki.org/allrecent should be updated to return the theme of the decks.
 
Also in the platform, the store AttachSubdeckStore should be modified to store the theme od the decks when it load the data of recent decks or searched decks.

In the platform, the service searchresults.lists should be revised in order to verify it returns the theme of the deck (I did'nt found it)

Sole-Valero:
Another thing, you now use:
 <Item.Image src={Microservices.file.uri + '/thumbnail/slide/' +deck.firstSlide+'/'+deckTheme} alt={deck.title} size="small"/> 

the url has not file extension...an also I didn't find the webservice thumbnail in the fileservice from experimental...do you have it in a local branch? In that case we can not test it...

kprist:
I forgot to include information about the other changes here as well, because I added those in the confluence page. So there are of course changes to both the deck service and the file service as I mentioned in the confluence page, and they have their own PRs:

https://github.com/slidewiki/deck-service/pull/120
https://github.com/slidewiki/file-service/pull/11

Sole-Valero:
@kprist, I show the changes in file-service. Also, I'll  try to see the changes in deckService, but I'm a liitle lost on them. However, I'm sure about it is also needed to update some services and actions used by the attach modals to retrieve the theme property. Now the code is incorrect. Also, it is needed to revert the change made in  AttachSubdeckStore about the content stored in deckSlides

kprist:
@Sole-Valero I understand what you're proposing about the deckSlidesThemes, but what I did works, we can change it how you proposed if it makes the code safer. The way I did it just included the minimum modification on the code. Also it was kindof strange that you had two separate arrays to store the id and the title of the slide, why couldn't we have a single array with all slide properties we might need grouped in an object ? 

First of all the updated deck service for /decktree/ and /deck/{id}/slides includes the theme as a slide property to make it easier for the platform code to figure out the slide to be applied to a slide, as a part of a deck.

I confirm that the /allrecent API needs a fix to include the theme as well, thanks for pointing it out.

The searchresults.lists does include the theme, I've added some code for that, but it needs the deck service to support that as well.

The url for the thumbnail does not have a file extension, it is not needed in general because it will still return a thumbnail image, I just wanted to make it more general to support other image types in the future, like png.

About testing the branch, it's a known problem that when we want to change multiple services it's not very easy to test, but I try to avoid merging stuff before proper review of all things. When I do such work or test such a work I just use local services with my local platform and a local mongodb. For this particular case, the fileservice APIs are new so we could in theory merge them, but the deck service changes cannot be merged without breaking experimental.

Sole-Valero:
@kprist I used it separated because also is used in other places for implementing accesibility. If it is modified, it should be broke the grid accesibility, for example.

/allrecent API needs a fix 

When it was possible, please send me an url to your server which contains the new service branches, so I can test it in depth.

kprist:
I have addressed comments by @Sole-Valero and also setup a server when you can actually test this branch, just override the URI's in the platform microservices.js file with the following:

```
'deck': { uri: 'http://slidewiki.imis.athena-innovation.gr:4000' },
'file': { uri: 'http://slidewiki.imis.athena-innovation.gr:4600' },
'search': { uri: 'http://slidewiki.imis.athena-innovation.gr:5000' },
```

I've also included an override for the search service so that tests that use search can be in sync with whatever is on the ATHENA deck service database. The database holds a recent dump from experimental, so same users and decks should exists there. The search index is almost empty, so to get results in the search pages please add new decks/slides or update existing ones so they can be incrementally included in the search index as well. 

Also, the thumbnails in the service are not pre-generated, so it might take a while for the images to display as they are created on-demand (about 1 sec per slide).

abijames:
@kprist 
Changing themes now correctly regenerates thumbnails. Subdecks are not affected. 

Noted one issue which needs to be consider for a separate branch. On some decks slide thumbnails are larger than slide content. I think this is realted to the slide aspect ratio e.g.  http://localhost:3000/deck/2654-1

PR #671, opened 2017-12-08
Swik 1915 translate ui of header dropdown

dpaun:
Internationalisation of the HeaderDropdown component


kadevgraaf:
Works well. Thanks!

dpaun:
I used context.intl because of aria-label encodings (didn't know how else to encode these).

DaryaTarasowa:
Ah, now I see. Makes sense. Then I think it is the best way. =)

PR #679, opened 2017-12-13
Allow empty description without replacing with 'empty' when not set

kprist:


kprist:
I've fixed the deck and import services to allow empty value for the description. It seems the deck metadata in general already contain decks with _truly_ empty descriptions, whereas others have the value 'empty'. This value was being added by the platform code when the users left the text field empty. 

With this PR this is no longer the case. I've also made some changes to the UI where it didn't handle empty descriptions correctly. However decks with 'empty' as description will still show that value, as it is a legitimate string value. We can make a database-level update of all the decks replacing 'empty' with '' (empty string).

abijames:
@kprist not sure if I am testing this correctly as its difficult to tell which decks would be changed (as many of mine still have empty displayed). But /deck/266 is not longer showing a description. Is that the expected behaviour?
Can you also confirm that new decks will no longer have "empty" for the description if the user decides not to enter anything?

abijames:
@kprist when I try to add a description to deck 266 I get an error: Error
Unknown error while saving. This also happens when I only change the title and do not change the description. I replicated it on other decks where the description was empty.

kprist:
Decks showing no description when there's none was intended on my part, both in deck view and deck listings. However we can change it however we like, now that actually can tell that a description is missing (instead of having an arbitrary 'empty' value).

Yes whenever user omits the description it should never get that 'empty' value stored in the database, whether they create a new deck, add a subdeck or edit the deck metadata. They can also remove the description by emptying the input text field. 

About the error you experienced, I think it's a preexisting bug you found, I also replicated it and the error returned is that the 'theme' attribute for the deck is empty. Maybe it was introduced after merging the themes PR, or it was there before it for those decks that lacked description, I am not sure, I need to investigate further. In the mean time for this PR please test again and if an error occurs trying selecting a theme from the dropdown to make it work.

kprist:
@abijames verified that it is a theme-related bug introduced with the themes PR (specifically the deck service changes ) and only manifests for some decks where the theme attribute is missing. Created JIRA task https://slidewiki.atlassian.net/browse/SWIK-1961 for this. Please test this by removing and re-adding descriptions from other decks preferably newly created ones.

kadevgraaf:
Works well, thanks!

PR #681, opened 2017-12-15
SWIK-1925 Internationalisation

rmeissn:
Wasn't able to test the changes quickly, as experimental seems to be down. Waiting for it to come online again in order to test that everything still works.

Please don't merge until someone verified these changes!


TBoonX:
Code worked.
Is the task really just about this File or about the whole "user Profile"?

rmeissn:
There were several tasks about specific files of the user-profile. Thus I guess it's only about this file.

PR #682, opened 2017-12-18
SWIK-1920 Internationalisation

rmeissn:
Please test and review these changes - social providers are not testable on localhost.

### DIFF
--- a/components/Login/ResetPassword.js
+++ b/components/Login/ResetPassword.js
@@ -7,11 +7,34 @@ import {publicRecaptchaKey} from '../../configs/general';
 import resetPassword from '../../actions/user/resetPassword';
 import resetPasswordResetStore from '../../actions/user/resetPasswordResetStore';
 import ResetPasswordStore from '../../stores/ResetPasswordStore';
+import { FormattedMessage, defineMessages } from 'react-intl';
 import ReCAPTCHA from 'react-google-recaptcha';
 import common from '../../common';
 
 class ResetPassword extends React.Component {
     componentDidMount() {
+        const messages = defineMessages({
+            mailprompt: {
+                id: 'resetPassword.mailprompt',
+                defaultMessage: 'Please enter your email address',
+            },
+            mailprompt2: {
+                id: 'resetPassword.mailprompt2',
+                defaultMessage: 'Please enter a valid email address',
+            },
+            mailreprompt: {
+                id: 'resetPassword.mailreprompt',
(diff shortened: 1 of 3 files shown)


components/User/UserProfile/Integrations.js
```diff
@@ -338,15 +413,15 @@ class Integrations extends React.Component {
                           <div className="two wide column">
                             {
                               (github) ? (
-                                <button className={github_disable_classes} name="github" onClick={this.handleDisable.bind(this)} >Disable</button>
+                                <button className={github_disable_classes} name="github" onClick={this.handleDisable.bind(this)} ><FormattedMessage id='Integration.disableGithub' defaultMessage='Disable'/></button>
                               ) : (
-                                <button className={github_enable_classes} name="github" onClick={this.handleEnable.bind(this)} >Enable</button>
+                                <button className={github_enable_classes} name="github" onClick={this.handleEnable.bind(this)} ><FormattedMessage id='Integration.enableGithub' defaultMessage='Enable'/></button>
```
TBoonX: Somehow this breaks the event transmitted to the function. On master the button is working.

components/Login/ResetPassword.js
```diff
@@ -56,12 +79,46 @@ class ResetPassword extends React.Component {
     }
 
     componentDidUpdate() {
+        const messages = defineMessages({
```
TBoonX: Why do you initialize this static content on each function call? Initializing it in the constructor and storing under this is what I would suggest.

TBoonX: Why do you initialize this static content on each function call? Initializing it in the constructor and storing under this is what I would suggest.

components/User/UserProfile/Integrations.js
TBoonX: Somehow this breaks the event transmitted to the function. On master the button is working.

components/Login/ResetPassword.js
rmeissn: Never tested storing the variable in "this.state"...just oriented myself as of the provided examples from confluence (where this is also initialised in functions).

TBoonX: Hm OK. No need to change it.

components/User/UserProfile/Integrations.js
rmeissn: Fixed by recent commit

PR #687, opened 2017-12-21
Removed CSS transformation to make the slide zoomable when it is disp…

vkovtun:
…layed on the view tab.


abijames:
@vkovtun I have had a quick look at your branch and chatted with @kadevgraaf ....

A simple horizontal scroll bar between the slide and the speaker notes would be a quick solution to make the work you've done complete. 

In the future, we could also look at going into the expanded view (hiding the panels on the left and right) when the browser is zoomed as we need to implement this for small screens as well.

kadevgraaf:
Hi Viktor.
Thanks for the changes!
Already looking good.
I do miss the initial resize for slide content to fit the browser window - for example: large slides are now loaded with half their content hidden (need to scroll to see other half)
Another nice feature that is now missing is that you see the borders of slide content - the double line in screenshot below.  This allows you to see whether slide content is outside the slide area - this is especially important for checking if something is shown in the top and bottom of the slide + it helps when you include an image that is shown only half/partial in the slide area.

screen shot 2018-02-07 at 10 07 53

abijames:
@vkovtun  @kadevgraaf  - looks good! I notice that this is only implemented in the slide view, not the slide edit yet. Is that deliberate?

vkovtun:
@abijames I had some problems with the edit mode, so I submitted my changes for the view mode for now. I also had to switch to more high priority tasks. I should come back to this one soon.

kadevgraaf:
Restarted from https://github.com/slidewiki/slidewiki-platform/tree/SWIK-1489---Slide-content-is-not-adjusting-to-brower-zoom---restart---vertical-space-and-zoomin-out-buttons

PR #688, opened 2017-12-21
SWAQ-575

rmeissn:
Exchanged button text and made it more reputable


kadevgraaf:
Thanks! (although I am a bit sad that my 👍 👍 icons are now gone :P )

PR #691, opened 2017-12-24
Added slide size option

kadevgraaf:


kadevgraaf:
TODO: have good thumbnails for previewing slide size

PR #692, opened 2017-12-26
Swik-1133

Sole-Valero:
This branch modifies ContentActionsHeader. 

Tab menu converted into an standard menu.
View and Edit tabs are now buttons:
- They are always visible
- view button is enabled in edit mode
- edit button is enables in view mode

All the buttons are accesible by tab key, but buttons are not accesible when they are disabled.

ContentActionsHeader uses Intl package to translate the text of the buttons.

String refs removed (we can't use them as are legacy)

This branch can be checked in my server: http://svalero.dsic.upv.es:3000


kadevgraaf:
Works very well. Thanks!
@abijames  - I am ok with merging this.

abijames:
i have made some changes to the branch to sort out some styling issues. My one remaining issue is that I am currently trying to style the Edit button as a basic button but the Save/Cancel button as standard buttons. But currently Edit button does not have a border and i can't fix it!

Some issues remain that need to be addressed but i think many of these are to do with the slide editor changes as well:

# when you expand the screen then no content is shown. This is a bug introduced by the new slide editor (SWIK-1996) but if we remove the 
# when you are on deckedit and select a slide you go to slide edit when expected behaviour would be to go to slideview. This is an exiting bug that I have logged at SWIK-1995

kadevgraaf:
I think I fixed the border by adding "attached" to the classnames. 
The other 2 issues are not related to the changes in this PR.
Are you okay with merging this version?

PR #693, opened 2017-12-27
SWIK-1934 Licence field removed from search forms in attach deck and …

Sole-Valero:
…slides modals

In order to check, attach slides or deck should be selected. Then, select the tab "From Slide Wiki" , and check the  the dropdown licence doesn't appear. Also, ensure it is possible making a search using the other fields without problems.


kadevgraaf:
Works well. Thanks!

PR #702, opened 2018-01-10
Swik 1989   terms and conditions are not open in a new tab from add deck page

kadevgraaf:


kadevgraaf:
there are multipe PRs to r2.7.0 - I will merge in 1 PR from master to r2.7.0

PR #706, opened 2018-01-10
Swik 2001   difference in image height between slide view and slide edit

kadevgraaf:


kadevgraaf:
there are multipe PRs to r2.7.0 - I will merge in 1 PR from master to r2.7.0

PR #708, opened 2018-01-11
Deactivated the load event of the iframe after listeners have been registered

rmeissn:


TBoonX:
Why?

rmeissn:
Because listeners were attached on each load event of the iframe. Load event occurred on each slide change. So: on each slide change, 3 new listeners are attached

PR #717, opened 2018-01-17
SWIK-2005_redirect_to_profilepage_after_signup

TBoonX:
Added navigation to user profile after social sign up


kadevgraaf:
Thanks!

PR #720, opened 2018-01-19
Swik 1814 select file button accesibility

Sole-Valero:
Select File and Add Deck button now respond when users press enter on them.
Branch available at http://svalero.dsic.upv.es:3000/


abijames:
@kadevgraaf I've testing this, it is ok to merge.

PR #722, opened 2018-01-22
SWIK-1106 Visual Deck Tree

rmeissn:
I've implemented a version of a visual deck tree that orientates at @abijames mockups.
Can you please review that the deck trees specific functionalities are still usable and that this matches your envisioned style?

I am currently not able to implement the accordion approach without the help of the people that implemented the deck-tree in the first place. I also recommend to refactor the deck-tree components to ease them and to get rid of unused functionality.

### DIFF
--- a/components/Deck/TreePanel/TreeNode.js
+++ b/components/Deck/TreePanel/TreeNode.js
@@ -13,6 +13,7 @@ import AttachSlides from '../ContentPanel/AttachSubdeck/AttachSlidesModal';
 import {connectToStores} from 'fluxible-addons-react';
 import ContentStore from '../../../stores/ContentStore';
 import {Button, Popup, Icon} from 'semantic-ui-react';
+import {Microservices} from '../../../configs/microservices';
 
 
 const findAllDescendants = (node) => Immutable.Set.of(node).union(node.get('children') ? node.get('children').flatMap(findAllDescendants) : Immutable.List());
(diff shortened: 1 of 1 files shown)


abijames:
I have added notes from reviewing this PR and subsequent discussion with @rmeissn to the JIRA ticket https://slidewiki.atlassian.net/browse/SWIK-1106

rmeissn:
Working on the proposed changes.
@kadevgraaf Do you think a temporary image (replaced by thumbnail) is needed? Thumbnail creation is pretty fast (and only required if the file is missing).

I may be able to implement 

> Another nice to have would be that subdecks have a slightly different outline colour that alternates per subdeck (now all is white, and the indentation for a subdeck is not very clear)

but will wait for @abijames input (see Jira Ticket)

abijames:
Tested lastest branch. All looking good. Aware Roy is continuing to work on focus indicator etc

kadevgraaf:
Ah indeed the thumbs are loaded really fast once generated - so indeed a temporary image is not necessary. Looking good!

rmeissn:
I have improved the keyboard selection styling of cards, as requested by @abijames . Is this okay/enough?

So we only miss the proper subdeck indication/styling, right? @abijames proposed to introduce a margin-left to subdecks, but I think that is only partly suited as we can have subdecks in a subdeck in a subdeck in a deck --> will result in very very small thumbs or a horizontal scrolling bar (that I do not prefer).
I think coloring subdecks (as @kadevgraaf proposed) is the better approach. E.g. by having a light grey background or overlay for the first subdeck and turn darker for each additional subdeck level. Nevertheless the background of thumbs (at least for the default theme) will always be white.

abijames:
@rmeissn do you have the html for what you propose for the styling of the list. I can try it out by adding it in the DOM. 
re sub-sub decks, if we have a clear indication of where sub (sub) decks start and finish then these should be clear

rmeissn:
These are two approaches: first slide is normal, second slide has an overlay (that may turn darker/lighter for different levels of subdecks), third slide was resized and got a background color (that may turn darker/lighter for different levels of subdecks).
example

rmeissn:
I've added a simple toggle button to switch between the two deck tree modes (I do not save it yet).

abijames:
Testing tree with the toggle-switch beneath it:

Switch works well but I can no longer access the deck tree using the keyboard. Currently when I tag from the fork button it goes to the toggle button. Expected behaviour is to tab to the first slide (then use arrows to move up and down), second tab would take focus to the toggle switch.

abijames:
On retesting the focus issue is not related to the visual deck tree. There are 2 accessibility issues that need to be resolved and then I am happy for this to be merged:

* Please can you remove H3 tags from around the thumbnails in the decktree. They cause navigation issues for screen readers.
* "Show thumbnails" label needs to be linked to the switch button using an id on the input and for on the label

Also, do we want to add (beta) to the Show thumbnails label to indicate this is still in development?

rmeissn:
@abijames Implemented. Can you please test the label thing again? Tried to do it with React Semantic Ui, but that's a little different than semantic-ui and I'm not sure this will work.
@kadevgraaf Fixed

abijames:
Toggle button now correctly labelled. This is ok to Merge!

PR #723, opened 2018-01-23
Swik 2012 fix edit question button visibiilty

dpaun:
This is a very small fix for the question edit button visibility (permission).


kadevgraaf:
Thanks @dpaun  !
Already works nice for correctly showing the edit-button.
Currently when I edit a question I just added and press "save" nothing happens. Are you able to reproduce this?

dpaun:
Yes, I'm able to reproduce it. Thanks! It's a bug, I'll fix it soon.

dpaun:
The problem with editing a question should be fixed now.

kadevgraaf:
Works like a charm. Thanks!

PR #725, opened 2018-01-23
SWIK-2023_titles_of_user_pages

TBoonX:
Added changing of browser tab title in actions.
Also fixed two tiny issues: public profile had classname instead of className and with new groups the name and description had the standard value "undefined"


kadevgraaf:
Thanks!

PR #728, opened 2018-01-25
SWIK-1936 link presentation to deck

huwf:
Adds a URL in the presentation back to the deck in a new tab.


kadevgraaf:
Hi @huwf 
Thanks for the work and PR!
I checked on my localhost - If I go to presentation mode of deck http://localhost:3000/deck/3859-1/slide/32595-1/32595-1:1/view to http://localhost:3000/Presentation/3859-1/3859-1/32595-1 
I get the following in my browser console: "Error: Unable to find element with ID 212."
It does seem to use the correct selector:
```
selector:  
Object { id: "3859-1", subdeck: "3859-1", sid: "32595-1" }
```

huwf:
This is a weird bug, lucky you tried it on that deck.  I don't know what's different about that, or indeed the specific slide it's on, since it's half way through. Interestingly, that slide doesn't seem to load properly in slide  view on experimental either.  https://platform.experimental.slidewiki.org/deck/3859-1/slide/32611-1/32611-1:17/view

kadevgraaf:
It might be because there is an error on the slide itself, so unrelated to the changes. (but it is hard to check since the slide is not loading :) )

abijames:
Just tested with a few decks and it worked ok. When a slide is empty the link does appear in the middle of the screen which isn't great. 

I am wondering if we customise https://github.com/denehyg/reveal.js-toolbar and add this link as there have been negative userfeedback that it is impossible to know the short cut keys. The toolbar created by the plugin is not accessible but I think we can fix this

kadevgraaf:
@abijames  and @huwf 

I do not get the error anymore, so that is fixed.

W.r.t. positioning - perhaps should it below/near the left/right navigation buttons? - otherwise I am not sure if everyone is happy with this (as you always get text in a presentation, even though you do not need it (I know how to close presentation mode / go back to deck (previous tab))

Shall we make a separate task/issue for the toolbar?

kadevgraaf:
This needs some more design work.
customise https://github.com/denehyg/reveal.js-toolbar and add this link might be a better solution as the link is now visible in presentation mode (which I think is not what presenters would want)

PR #729, opened 2018-01-25
Fixes SWIK-2027

rmeissn:


kadevgraaf:
keyboard access is not working for me in Firefox.
Perhaps it requires a surrounding div or button element instead.

rmeissn:
I've exchanged the i element to a proper button. Tested on Chrome and Firefox: everything can be traversed by tabs.

I had to exchange the default colors of both buttons to proper "red" and "black", as their default colors ("github" and "google plus") did not change on hover/selection. Normal colors show a highlighted style.

One flaw I noticed: If I hit tab after "close" was focused, I am able to traverse the rest of the site by hitting tab. I guess this is not supposed to happen?

abijames:
@kadevgraaf @rmeissn all ok now! Good to merge

PR #730, opened 2018-01-25
Swik 1848 tag recommendation

schatzopoulos:
Adds recommendations for tags in the tag panel of each deck

-- The nlpstore and nlpservice need to be deployed in order to show tag suggestions


kadevgraaf:
Cool. Works very nicely!

@abijames  - the add/dismiss buttons probably need keyboard-access/tabindex improvements and aria-labels?

W.r.t. user experience, I would expect that the tag gets added when I click on it, but instead I go to deckfamily. I think this is OK/minor, but maybe it is a quick fix to have a link to deckfamilies ('see decks tagged by this keyword') and then have the add onclick on the tag + an x-button to dismiss the tag suggestion.
Another possible improvement might be the number of suggestions - currently 4 tags are suggested and on dissmiss a new one is shown - perhaps this could be 8 or 12 - seeing more tags also gives you a better idea/overview of which tags are appropriate/suitable and which are less appropriate/suitable.

Also I was wondering when deckfamilies get updated/indexed? - If I e.g. add tag citizen-science to http://localhost:3000/deck/3859-1 and navigate to http://localhost:3000/deckfamily/citizen-science the deck is not yet show. Is this solr re-indexing e.g., every hour?

schatzopoulos:
Re: indexing of deck families - the index is updated every second but it seems to be an error now on experimental. @bwulff can you restart solr and search-service on experimental?

abijames:
Comments made on ticket https://slidewiki.atlassian.net/browse/SWIK-1848

abijames:
Recommendations to help with the card resize issue made on JIRA ticket

kadevgraaf:
Works very well now. Thanks!

PR #731, opened 2018-01-25
SWIK-1771-Fix spellcheck suggestion redirection on firefox

schatzopoulos:
Fixes issue that was reported in SWIK-1771, that search spellcheck suggestions are not properly redirecting in firefox.


kadevgraaf:
Thanks!

PR #732, opened 2018-01-25
Fixes SWIK-2028

rmeissn:
Added language chooser to mobile sidebar + fixed sidebar not closing on some click targets


abijames:
Works well. Only comment for @rmeissn is that the Greek flag is missing. Once merged into master this should be checked again on a mobile device.

kadevgraaf:
Thanks!

PR #727, opened 2018-01-24
SWIK-1692 social share buttons in presentation rooms

rmeissn:
I've implemented social share buttons for presentation rooms. URL sharing has been also moved to this menu. I've tried to add useful aria attributes. Additionally a twitter hashtag button has been added, that uses an autogenerated hashtag.

Is it better to share a link to slidewiki view/edit mode instead of the presentation mode?

Sharing via E-Mail is unfortunately not possible due to a browser bug of chrome (at least I think it is). I've reported this bug at [the chromium bug tracker](https://bugs.chromium.org/p/chromium/issues/detail?id=804851&q=component%3ABlink%3EWebRTC%20&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified)

This PR does not resolve Ticket SWIK-1692. It's just a part of it.

social buttons

PR #737, opened 2018-01-29
SWIK-1692 Twitter timeline in presenation rooms chat

rmeissn:
I've included twitter tweets to the chat of the presenter and also all listeners (peers) of a room. A room streams new tweets for the hashtag of the room (e.g. "#SWORG #D15R5" (deck 15-6) to the presenter (from the signaling service). The presenter than relays these tweets to all of his peers.
Tweets can be excluded from the chat by a checkbox and the "clear chat" button removes all tweets too.

@abijames Tweets look very different than normal chat messages, cause I'm currently using the tweet embed from twitter itself. Is this okay (and accessible) or should we implement our own tweet component that uses react-semantic-ui only?
See this Screenshot: 
tweet timeline showcase

I've noticed that for a larger number of tweets (>20, with pictures, links, ...), enabling to show tweets again feels a bit laggy.

This PR needs to be tested with [this branch of the signaling service](https://github.com/slidewiki/WebRTC-Signaling-Service/pull/4) (EDIT: already deployed to exp)
Social Share buttons (incl. tweet) is implemented in [this branch](https://github.com/slidewiki/slidewiki-platform/pull/727)


rmeissn:
@ali1k Can you please have a look at the code, functionality and also linked PR of the signaling-service?
Some questions:
Should we use Twitter embeds (like currently implemented) or implement custom boxes for Twitter messages? Using Twitter embeds enforces to integrate Twitter frameworks, that includes user tracking (that I am not a fan of, cause of privacy thoughts!).
Can you verify that it works as it should? Kurt and me are not Twitter users....
I'm really not sure if I implemented https://github.com/slidewiki/WebRTC-Signaling-Service/pull/4 (already merged) right, as the Twitter API is a bit crappy if it comes to streaming. Is it okay to open a new stream for each room or must this be implemented differently?

rmeissn:
Please see this news article about copyright of embedding tweets: https://www.golem.de/news/us-gerichtsurteil-einbetten-von-tweets-kann-urheberrecht-verletzen-1802-132838.html
Titel translated to English: "Embedding Tweets may break copyright" --> As the article is in German, this may help finding an English one.

rmeissn:
@kadevgraaf Do we want to wait for someone familiar with Twitter to review this PR or is it fine to merge it? @TBoonX seemed to be pleased with the quality.

ali1k:
Sorry for my late review. I had overlooked this PR and just knew about that by Klaas. First of all, the copyright issue doesn't yet seem very serious. It was on a very specific case. I read the details in https://www.theverge.com/2018/2/16/17020278/tweet-embed-copyright-infringement-justin-goldman-tom-brady-photo-ruling and https://www.engadget.com/2018/02/16/tweet-embed-copyright-violation-judge-ruling/ I don't think we need to do something with that at the moment. If it gets more serious we can then think about other methods for embedding of tweets without displaying the images, etc.

I tried to test the feature but cannot get it to work on my machine. I fixed a minor variable typo but still receive an error at line 42 of presentationBroadcast.js:

`this.deckID = this.props.currentRoute.query.presentation.toLowerCase().....`

@rmeissn can you check that?

rmeissn:
@ali1k Something broke PRs on my local environment and also on experimental. Unfortunately I am unable to interpret the error message. This is thus not related to this PR. Will post on Slack.

rmeissn:
@ali1k Fixed it. It's now working again.

ali1k:
great! I tested it again with some manually inserted hash tags in the presentationBroadcast.js file.  It seems to work well. Just some suggestions:
- as far as I understand any tweet which has both '#SWORG' and '#DechIDRrevID' will be streamed. shall we use #SlideWiki instead of #SWORG
- in my opinion, the new tweets (or questions) should come to the top of the list; or it should should automatically scroll down to the new tweets coming up, otherwise, the presenter need to manually scroll down to see the new content appearing there.
- do we have a threshed for the number of tweets that can be listed there? IMO, would be good to remove or archive the old tweets if the list get's ver long.

rmeissn:
Many thanks!
About your points:
1. This is how experienced the twitter streaming API works ^^ My background thoughts were: #SWORG is a little shorter than #SlideWiki and might be better suited for future private/other instances of SlideWiki. That's just a proposal for the base hashtag....if there is a better one, we should use it.
2. That's exactly what a university teacher in Leipzig appreciated: Scroll to some entry and discuss it with students, without having the panel automatically scrolled to new messages/tweets. I thought about implementing autoscroll + on/off button, but didn't found a quick solution for autoscroll.
3. There is currently not threshold, but a button to discard the chat (will also discard tweets).

ali1k:
I think for the moment it is good enough to be merged. Maybe we can have another PR later on to address the minor issues discussed:
1. #SlideWiki in my opinion is better than #SWORG for branding (with current extension of Tweet limits, the size should not be an issue). It took me some time to find out SWORG refers to SlideWiki.org Maybe @abijames has also a feedback on this.
2. There are different ways to handle upcoming tweets. We can use the same design pattern as twitter to show the number of upcoming tweets and when user clicks on that sees the new tweets, or we can autoscroll to the new ones. For auto scrolling you can simply use
```javascript
$('html, body').animate({
    scrollTop: $("#target-element").offset().top
}, 1000);
```
3. I think it would be still good to have threshold (e.g. max 1000) to avoid the browser get stuck in case lots of tweets are streamed.

PR #738, opened 2018-01-29
Added enlarge button to presentation room

rmeissn:
Presenters and Listeners now have the ability to enlarge the presentation without knowing the shortcut of reveal.js. This has been requested by some people at our university.


kadevgraaf:
Nice/good feature addition!
Works well in Chrome

In Firefox Dev I get the following console message:
"Request for fullscreen was denied because at least one of the document’s containing elements is not an iframe or does not have an “allowfullscreen” attribute.". Perhaps an allowfullscreen attribute could be added dynamically / on-load

Does not seem to work in Safari: "An error occured
We're sorry, but we can't connect you to the presenter. It seems like there is a problem with your connection or browser. Please update your browser, disable extensions or ask your network operator about it. We're using a peer to peer connection technique called WebRTC." 
- but this is a different problem it seems

rmeissn:
Fixed the FireFox problem.

I can't test for Safari (no Mac). Was the presenter tab still open? Please test with a different room name (e.g. test3).

Do you have a recommendation for a tabIndex? Never used these before....

kadevgraaf:
Works well now, thanks!

PR #745, opened 2018-01-30
Add translation strings to search components

schatzopoulos:
translations strings for search components according to SWIK-1924


kadevgraaf:
Thanks!

PR #746, opened 2018-01-30
SWIK-2036-fixDeleteButton deletebutton: same condition to disable and…

Sole-Valero:
… tabindex -1

The condition checked to asign -1 to the tbaIndex of the delete button is now exactly the same used to disable the button. Thus, I hope the problem was fixed now.


abijames:
+1 @kadevgraaf this is ok to merge.

PR #750, opened 2018-02-06
Swik 2021 fix import modal

Sole-Valero:
- Modal changed to react-modal, in order to solve accessibility issues.
- color of the button updated following the instructions (blue and red)
- If users select a wrong input file, an swall message inform about it
- Modal can be opened and close, and buttons still work.
- If users cancel, no change is introduced
This branch is in my server http:// svalero.dsic.upv.es:3000


Sole-Valero:
@abijames , actually the focus goes to the upload button..as it is the following step...

abijames:
@kadevgraaf  this is ok to merge

kadevgraaf:
Thanks!

PR #753, opened 2018-02-08
improve page title by only including deck title and slide title.

pierorex:
Before, it was like this: "SlideWiki | Deck | some number | slide | even more complicated number | view"
Now, it is similar to this: "SlideWiki | <Deck Title> | <Slide Title>".

I had to kind of hack the store that handles the page title to avoid getting overridden by other modules that also update the page title and are executed after the loading of decks and slides.
Is this solution acceptable?


stores/ApplicationStore.js
```diff
@@ -5,11 +5,15 @@ class ApplicationStore extends BaseStore {
     constructor(dispatcher) {
         super(dispatcher);
         this.pageTitle = '';
+        this.frozen = false;
     }
     handlePageTitle(payload) {
         this.dispatcher.waitFor(RouteStore, () => {
-            this.pageTitle = payload.pageTitle;
-            this.emitChange();
+            if (!this.frozen || payload.allowUnfreeze) {
```
kadevgraaf: Can anyone else also take a look at this? seems a bit tricky, but if it does the job I think it is good.

pierorex:
Sure, is there an issue for that or should I just add it to this branch?

pierorex:
Well, I already started working on the same branch

pierorex:
Can someone check/merge this?
@kadevgraaf @bwulff

kadevgraaf:
hi @pierorex  - thanks for the additions and sorry for my late reply. I hope to look at this PR soon.

kadevgraaf:
I think this somehow breaks the fluxible/store progress. 
I tried switching between viewing and editing slides, but then there are strange inconsistencies and parts of the UI that are not updated.
Can you try testing this and see if there is a cause this these inconsistencies? It might also have to do with changes that were made in the mean time.

pierorex:
I'm checking it and I don't find any UI issues, could you tell me exactly what you're experiencing so I can focus on that?

pierorex:
I just saw one issue: when in 'Edit' mode, if I try to use the arrows to move between slides the slide area gets buggy. I'll push a commit now that fixes that and hopefully also fixes what you were seeing.

kadevgraaf:
Thanks for the fix! seems to work now 👍

kadevgraaf:
Maybe one more improvement would be to reset the page title once you exit a slide/deck view/presentation.
When testing I would first view slide "name XYZ", and the title correctly reflected that, but then when I viewed my decks page (e.g., http://localhost:3000/user/kadevgraaf), the title remained to be "name XYZ"

kadevgraaf:
Hi @pierorex 
I removed unnecessary (and perhaps also some necessary) page title updates from /actions (e.g., from actions/tags/loadTags.js).

By doing this we do not need the freeze and allowUnFreeze anymore, and we might improve performance because of less page title updates.

Can you check if there are necessary page title updates that I commented out by accident? E.g. user profile and my decks might need page title update that are triggered from /actions. 

Also, some slide titles have HTML in it - e.g. http://localhost:3000/deck/3825-1/slide/177-5/177-5:2/view and http://localhost:3000/deck/3825-1/slide/178-4/178-4:4/view  -> this does get removed from the breadcrumb (top of decktree) you could reuse this code.

kadevgraaf:
Works well. Thanks!

PR #754, opened 2018-02-09
Swik 1827 loading indicator for info panel

vkovtun:


kadevgraaf:
Thanks for the performance improvements!
Do you also want to add a loading indicator similar to, e.g., slide view panel in components/Deck/ContentPanel/SlideModes/SlideViewPanel/SlideContentView.js 
Using, e.g.,
`{(this.loading === 'loading') ? <div className="ui active dimmer"><div className="ui text loader">Loading</div></div> : ''}`
and 
```
    componentWillReceiveProps(nextProps){
[..]
        if (nextProps.loadingIndicator !== this.props.loadingIndicator)
        {
            if (nextProps.loadingIndicator === 'true')
                this.loading = 'loading';
            if(nextProps.loadingIndicator === 'false')
                this.loading = '';
        }
}
```

vkovtun:
@kadevgraaf I tried to do that, but it turned out to be not as easy as in the case of the slide view. I will try your suggestion.

vkovtun:
@kadevgraaf I forgot to answer your last question. I had to use that function to improve performance. The application has to have an equals method for objects to be able to see whether there have been any changes or not. To the best of my knowledge there is no native function to accomplish that. Another alternative would be `JSON.stringify(obj1) === JSON.stringify(obj2)`, but it suffers from false negatives, since the following objects would not be equal: `obj1 = {a: 1, b: 2}` & `obj2 = {b: 2, a: 1}`.

vkovtun:
@kadevgraaf I have implemented the loading indicator.

kadevgraaf:
Thanks! Works nicely.
1 minor thing - When I view a slide, e.g., http://localhost:3000/deck/4047-1/slide/335-3/335-3:12/view , and I refresh the page using ctrl-R then the first time the info panel keeps loading. The second ctrl-R refresh it does work however. If this is too troublesome to fix I think we can still merge.

vkovtun:
@kadevgraaf I cannot reproduce your problem. Can you please check it again? If you can reproduce it, maybe you can send me a link to the slide, where this happens?

vkovtun:
@kadevgraaf I merged master into this branch, fixed the conflicts and merging bugs. Still I cannot reproduce the issue you have reported.

PR #755, opened 2018-02-09
Fix access to collection page when not signed in

schatzopoulos:
Fixes issue that one cannot view a deck collection page (eg http://localhost:3000/collection/1?sort=order) when not signed in


kadevgraaf:
Works well compared to current master. Thanks!

PR #756, opened 2018-02-09
Swik 2058 bug fix

Sole-Valero:
Recent ImportModal needs an extra state to control it is showing an error to the user.
This fix adds it in order to prevent possible problems. Also, it changes the way we indicate which colors we want to use for buttons, in order to avoid some warnings.

### DIFF
--- a/components/Import/ImportModal.js
+++ b/components/Import/ImportModal.js
@@ -281,7 +281,7 @@ class Import extends React.Component {
                                         <p>{this.context.intl.formatMessage(this.messages.modal_explanation2)}{MAX_FILESIZE_MB}MB).</p>
                                       </div>
                                       <div className="ui input file focus animated">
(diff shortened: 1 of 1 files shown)


components/Import/ImportModal.js
```diff
@@ -107,24 +108,22 @@ class Import extends React.Component {
...
-
-        if(!this.state.cancelled && !this.state.accepted){ //user clicks outside to cancel without pressing upload button
-            this.setState({
-                cancelled:true
-            });
-
-            if(this.props.ImportStore.fileReadyForUpload){
-                this.context.executeAction(importCanceled, {});
+        if(!this.state.onShowError){
+            if(!this.state.cancelled && !this.state.accepted){ //user clicks outside to cancel without pressing upload button
```
pierorex: For me, when clicking outside the import modal, it doesn't close it so I'm not able to check if this part works

Sole-Valero:
But if you press esc, you will exit.

abijames:
I have tested this and it is ok to merge @kadevgraaf

kadevgraaf:
Thanks!

PR #757, opened 2018-02-09
Improve image cropper previous state

kadevgraaf:


kadevgraaf:
@abijames  - when you test the modal text will still show "There it is. You catched it! - Seems like we can not handle" even though it is changed in code (this is due to, e.g., en.json not being updated yet).

kadevgraaf:
Works well, thanks!

PR #758, opened 2018-02-12
Swik 1908 add play button deck card

abijames:
Adds following to the deck card:
* Open and Play button. **Currently play button does not work**
* No of likes - **this needs to completed as likes number is hard coded**
* Shortens deck title so it fits on 1 line
* Adds headings and aria properties to improve accessibility


kadevgraaf:
Works well. Thanks!

PR #767, opened 2018-02-21
SWIK 2045 WCAG keyboardtrap

kadevgraaf:


abijames:
Works well. Fixed keyboard trap in the speaker notes in Chrome and FF with NVDA

kadevgraaf:
Thanks for the review!

PR #768, opened 2018-02-21
SWIK 2069  Slide edit  Speaker notes below slide only 2 lines visible. Please allow resizing this area

kadevgraaf:


abijames:
all good!

PR #770, opened 2018-02-22
SWIK 2043      Improve slide title change in slide editor

kadevgraaf:


abijames:
Some possible issues:
* when i click the new submit button then the slide content disappears 
image
* the placeholder text "slide name" needs to be a lighter font than the text entered
* would be nice to be able to press enter to submit
* can you add aria-label="submit slide name" to the submit button on the field
* message below box can be made clearer: "Your slide name will be updated once you have saved and exited the slide edit mode."

kadevgraaf:
Thanks for your feedback!
- I now made a SWAL modal to ask if the user wants to save the slide + name and exit slide edit mode, or continue editing.
- I reformatted the input area to be a textarea with semantic-ui property fluid, to fit the edit menu and to allow easier editing of long names.
- I switched some aria labels, also to prevent trigerring the button by clicking on text.
- Is the placeholder text browser specific or can I add a CSS trick for that? .For now I removed placeholder="Slide name"
- I updated the message below box - this still needs to be updated by TransiFex, so old message is still shown
- updated languages

I detected a minor bug - if slide name becomes too long, the layout is messed up: 
screen shot 2018-02-26 at 18 30 21

abijames:
A dew comments :-) Can you change the name of the button to start changing the slide name to "Change slide name"
I like the alert for saving the slide but needs a few tweaks:
* if we have the SWAL modal than the text below the edit name box is redundant and can be removed
* when the SWAL modal if loaded, focus needs to be placed in the modal (all other aspects of focus management are good)
* there doesn't need to be aria-labels on the yes/no buttons as this just repeats the text labels on the buttons.

kadevgraaf:
Thanks! All done except last point:

-  there doesn't need to be aria-labels on the yes/no buttons as this just repeats the text labels on the buttons.

I do not know where to change this. There is no mention of aria labels in the definition of the SWAL modal

kadevgraaf:
"Change slide name" is not updated in the left edit panel because default.json is not updated, but it should be good... :)

PR #774, opened 2018-02-26
Swik 2076 accessibility deck card

abijames:
Improved formatting of metadata, removed reducendent tabindex, added aria-label to deck title


abijames:
@kadevgraaf when I run this branch on localhost, I get an error on the My decks page before the deck cards appear. However I don't think this is my code changes as other pages work fine. So can you check the code before merging?

PR #776, opened 2018-02-27
Swik 1500 fix user menu fix user dropdown menu

sednanref2004:


abijames:
@sednanref2004 works great !!! @Sole-Valero @kadevgraaf can you test this both asap so we can try and get it into the release before the testing next week?

kadevgraaf:
Works great indeed.
In Firefox the menu options are not highlighted. Works well in Safari and Chrome.

PR #777, opened 2018-02-27
Put space HTML character in empty blocks to have view and edit the same

kadevgraaf:
config.fillEmptyBlocks = true; instead of false, to reset it to default
addition of &nbsp; in empty/newline blocks by CKeditor

### DIFF
--- a/assets/ckeditor_config.js
+++ b/assets/ckeditor_config.js
@@ -29,7 +29,7 @@ CKEDITOR.editorConfig = function( config ) {
 
     // Set the most common block elements.
     config.format_tags = 'p;h3;h4;h5;pre';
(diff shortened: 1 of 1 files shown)


kadevgraaf:
Compare view and edit mode between, e.g., platform.experimental.slidewiki.org/deck/3883/slide/32875-4/32875-4:1/edit with the original bug report slide: https://slidewiki.org/deck/107549-1/slide/703890-6/703890-6:3/view

This is a critical change I think -> if something is wrong then any slide that is saved might be affected.

abijames:
@kadevgraaf seems to work ok. However, experimental services are still very slow...

PR #780, opened 2018-02-28
Swik 2034 fixes for groupdetails

TBoonX:
DONE:

- modal should have aria attributes:
- role=dialog
- aria-labelled by the header
- **aria-described by the text in the modal
- ensure keyboard focus is trapped within the modal
- allow modal to close when ESC key pressed

Please try if your problem is still there:
_And.... when I open up deckedit after this bug issue the Group name for the groups that I have looked at the details the previosu time, does not appear in the list_


TBoonX:
If its alright we could add it to this weeks release

TBoonX:
The modal has still multiple issues. I will improve it and add the reviewers again.

abijames:
tested ok @kadevgraaf this is good to merge

PR #781, opened 2018-03-01
LTI Handler added

urashidmir:
LTI Handler added


urashidmir:
LTI Handler added

urashidmir:
LTI handler added.

PR #783, opened 2018-03-02
SWIK-1318 Implemented loading indicator on the edit page.

vkovtun:


kadevgraaf:
I have implemented your solution (thanks!) in combination with other fixes in https://github.com/slidewiki/slidewiki-platform/pull/835

PR #785, opened 2018-03-06
Changed mode on nodeTree so it goes to the view mode even when editin…

sednanref:
…g a deck.

As far as I've seen inside the code and the complex structure of the Deck Tree, this is the only change needed for this 'bug', since now during slide edition there's a new component 'slide editor' panel where previously we had the Deck Tree, so it's not necessary to have the previous 'mode' into consideration and just go directly to the slide-view.


kadevgraaf:
That is an elegant fix :) Thanks!

PR #788, opened 2018-03-06
initial local zoom (need to use user store)

kadevgraaf:


abijames:
Just testing this.... If the buttons are in a div "ui vertical icon buttons" then it will create a toolbar like this 
image

kadevgraaf:
@abijames  - we made improvements - can you check if this is ok? Then we can do the same in slide edit mode

abijames:
Testing 8/3.... When I zoom out to 25% the slide gets smaller and also there is no horizontal scrollbar. I would like to improve the icons on the toolbar but that can wait...

PR #790, opened 2018-03-08
snappy grids icanhas

kadevgraaf:


abijames:
Looks good....

At some point we need to think about keyboard controls for moving boxes. I think that will be easier with this grid lock. Somethin like alt and arrow keys when an input box is selected.

kadevgraaf:
Re: keyboard controls for moving boxes. I think that will be easier with this grid lock. Somethin like alt and arrow keys when an input box is selected.

We have this -> if you mouse-click-select/tab to an input box, you can move them around with control+alt+arrrow keys. (by providence/accident) it has the same grid-positioning of 10 pixels intervals of the current snappy grids config I made :)

Shall I merge this?

PR #792, opened 2018-03-08
SWIK-2092-fixDownloadModel 

Sole-Valero:
Now the dialog is closed after the user selecs an optionand clicks on the download button


kadevgraaf:
Works well. Thanks!

PR #794, opened 2018-03-09
SWIK 2085 keyboard trap presentation mode

kadevgraaf:


kadevgraaf:
TODO: I need to add an empty/hidden element at the start of each slide which gets focus

PR #797, opened 2018-03-13
SWIK-2111 fixing activity feed store update

dpaun:
Activity feed does not update properly when there is a new comment on a deck with id which has no explicit revision number. This a small fix for this issue.


kadevgraaf:
Thanks for the fix!

PR #799, opened 2018-03-13
SWIK-1982 Show mobile header on any screen where the user menu can't be shown due to the width of the browser

vkovtun:
Made MediaQuery take into account not the width of the device, but the width of the web page. This fixes the issue with not working Hamburger menu on desktop computers with modern screen resolution.


abijames:
Tested in Chrome, FF, Edge, IE
- works well on Chrome, FF
- on Edge, IE it is ok but there is one point where the user menu button can disappear before the header concatenates. I am happy to live with this and improve if we get user reports this is a problem

On my localhost, performance was slow @kadevgraaf can you check if you expereience this as well as it could just be my computer struggling generally today...

kadevgraaf:
Works well!
FF, Chrome, Safari.
@abijames  - runs fine/fast on my localhost.

PR #801, opened 2018-03-15
Add nlp to microservices template and SERVICE_URL_NLP to docker-compose

schatzopoulos:
I noticed that the platform cannot establish connection with the nlp service on experimental, so I added nlp service to microservices.js.template and the respective env var to docker-compose.yml


kadevgraaf:
Thanks for the fix!

PR #805, opened 2018-03-20
Fixed new zIndex to avoid transparency.

sednanref:


abijames:
@sednanref thanks works well. @kadevgraaf this is a bug fix and good to merge.

PR #812, opened 2018-03-23
Improve slug URL format; include slugs in even more links

kprist:
This puts the friendly URLs in more places (but not everywhere!), also addressing the core issue with SWIK-2135:
> URL for launching presentation mode from deckview does not contain readable slug

As a bonus, some `<NavLink>` elements have replaced some `<a>` elements, and the slide thumbs in the deck view no longer overlap with the slide title text.


kprist:
@rmeissn the fact you need to click twice for the back button to work is an old bug, try visiting any page then look at your history in browser in current version. I think I fixed it in this branch. I also merged master into this PR.

I verified the bug you described, but can't seem to figure out why it doesn't work. I noticed that if you first go to the home page then go to my decks, it works! If you go to my decks and reload the page (server side rendering), then the bug appears! Any ideas where to look for it ? In the console all I get is `Error: string is undefined`.

I also verified the same bug in all links where the title is included like in the featured section, not just my decks page.

PR #814, opened 2018-03-26
Swik 2065 fix sign in modal

Sole-Valero:
Missing aria-labels has been added to the sigin modal. Also, the modal content is surronded by the react trap component. In order to manage the trap, the modal has been connected to an store, and a new action to activate the trap is called when the modal is open. Also, the same action is called when the onHidden method occurs.

### DIFF
--- a/components/Login/LoginModal.js
+++ b/components/Login/LoginModal.js
@@ -16,6 +16,8 @@ import {FormattedMessage, defineMessages} from 'react-intl';
 import SelectInstanceModal from '../User/SelectInstanceModal.js';
 import openSSOModal from '../../actions/user/openSSOModal';
 import FocusTrap from 'focus-trap-react';
+import LoginModalStore from '../../stores/LoginModalStore';
+import updateTrap from '../../actions/loginModal/updateTrap';
 
 const headerStyle = {
     'textAlign': 'center'
@@ -37,14 +39,11 @@ class LoginModal extends React.Component {
         this.provider = '';
         this.isLoading = false;
         this.state = {
-            activeTrap: false,
+            activeTrap: this.props.LoginModalStore.activeTrap?this.props.LoginModalStore.activeTrap:false,
         };
 
         this.errorMessages = defineMessages({
-            headerText:{
-                id:'userSignIn.headerText',
-                defaultMessage:'Sign In'
-            },
+
             error403: {
                 id: 'userSignIn.errormessage.isSPAM',
                 defaultMessage: 'Your account was marked as SPAM thus you are not able to sign in. Contact us directly for reactivation.'
@@ -59,11 +58,18 @@ class LoginModal extends React.Component {
             }
         });
     }
+    componentWillReceiveProps(nextProps){
+        this.setState({
+            activeTrap: nextProps.LoginModalStore.activeTrap
+        });
+
+    }
     unmountTrap(){
         if(this.state.activeTrap){
-            this.setState({ activeTrap: false });
-            $('#app').attr('aria-hidden','false');
-        }
+            this.context.executeAction(updateTrap,{activeTrap:false});
+        };
+        $('#app').attr('aria-hidden','false');
+
 
     }
     isModalShown() {
@@ -73,11 +79,8 @@ class LoginModal extends React.Component {
 
     handleLoginButton() {
         $('.ui.login.modal').modal('toggle');
-        this.setState({
-            activeTrap: true,
-        });
-        //hidden the other page elements to readers
-        $('#app').attr('aria-hidden','true');
+
+
         setTimeout(() => {
             ReactDOM.findDOMNode(this.refs.email1).focus();
         }, 0);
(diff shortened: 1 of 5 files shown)


abijames:
Tested and all fine.... @kadevgraaf you can merge....

components/Login/LoginModal.js
```diff
@@ -31,11 +34,16 @@ class LoginModal extends React.Component {
         this.handleLoginButton = this.handleLoginButton.bind(this);
         this.handleSignupClick = this.handleSignupClick.bind(this);
         this.handleNoAccessClick = this.handleNoAccessClick.bind(this);
+        this.unmountTrap = this.unmountTrap.bind(this);
         this.signin = this.signin.bind(this);
         this.provider = '';
         this.isLoading = false;
+        this.state = {
+            activeTrap: this.props.LoginModalStore.activeTrap?this.props.LoginModalStore.activeTrap:false,
```
TBoonX: Is there a reason to use the state? Instead just use this.props.LoginModalStore.activeTrap wherever this.state.activeTrap is used?

Sole-Valero: It is needed to use state if you want mount/umount the trap. Changes in the state provoque it re-renders.

TBoonX: Re AddDeck: 
The bug is the following:
When I am not logged in and click on "Add Deck", the LoginModal opens. When I then close it, the modal closes but the underlaying view keeps black.
Suggestion:
Please update the action loadAddDeck in order to have there the same mechanism as in you already changed code.
In general: Please make sure the usage of the LoginModal is everywhere the same.

Sole-Valero: Now I undertand excactly the problem!! I only modified the code to open the modal from the sigin button, because I didn't know it was also used from another point!!. Thanks a lot for your review. I need toupdate also the call from the adddeck

Sole-Valero: and from deckEditPanel, ResetPassword, UserProfileReview, UserProfileReviewUser, UserRegistration

Sole-Valero:
@TBoonX  I think I patched the code in order LoginModal was open and close in the same way from all the places. Please, check it again from my server.
svalero.dsic.upv.es:3000

actions/loadAddDeck.js
```diff
@@ -10,7 +11,32 @@ export default function loadAddDeck(context, payload, done) {
...
         context.executeAction(navigateAction, {url: '/'});//, reqId: payload.navigate.reqId});
-        $('.ui.login.modal').modal('show');
+        //prepraring the modal
+        context.executeAction(updateTrap,{activeTrap:true});
+        //hidden the other page elements to readers
+        $('#app').attr('aria-hidden','true');
+
+        $('.ui.login.modal')
+        .modal({
+            onVisible: () => {
```
TBoonX: Why is the configuration needed here again? Header.js should only configure the modal and the other places which opens the modal should just open it.
Thus the following should be enough:
context.executeAction(updateTrap,{activeTrap:true});
 $('#app').attr('aria-hidden','true');
$('.ui.login.modal'). modal('show');

Sole-Valero: I think that the semantic ui has some problem when is called from an action, and not from a component. When the modal is open from a compoenent, you can see how the class of the body tag changes to dimmable dimmed. This not happens when it is shown from the action, and the dimmer is not hidden. If we add this class using jquery (as automatically occurs when you show the modal from a compoennt), when you close the modal, the dimmer is hidden. 
For this reason, I have commented this onVisible setting:
//This class is added automatically when the modal is called from a component, not an action.
 //If it is not set, the dimmer is not disabled when the modal is hidden

TBoonX: OK the reason should be that we have an isomorphic application (sometimes actions are executed on the server which have to initiate the modal again)
But if you configure the modal in Header.js you dont need to do that again in the other components, right?

Sole-Valero: I think you configure it each time you shows it...and in header is not needed to configure the onVisible...because it works ok.

Sole-Valero: and...in header the modal is only configured if ou clicks on SigIn button.

All these problems disappear if the nodal from react-semantic ui is used, because you control the modal using the same way than other components. But, I tried to patch this component using the current code, and transform it to a totally react comeponent, as it has several code....

TBoonX: I created a branch SWIK-2065-fixSignInModal_proposal in where the LoginModal is initiated just one time in the Header. Please check if this is interesting to be merged.

In every case please check that LoginModal uses itself in the preferred way. At the moment there are many calls in there which just do: $('.ui.login.modal').modal('toggle');

Sole-Valero: I think this hould be checked for the original programmer who choose use toggle instead...I have tried to undertand all the code, and I think than toggle was use to hide in all cases...but original programmers should check it.

Sole-Valero: @TBoonX  is going to finish it, now he has the ticket assigned

kadevgraaf:
Thanks @Sole-Valero and @TBoonX !

PR #815, opened 2018-03-28
Swik 2137 duplicate ids in edit mode

vkovtun:
I was also working on the issue https://slidewiki.atlassian.net/browse/SWIK-1951 therefore I reformatted some other lines in the code.

### DIFF
--- a/components/Deck/ContentPanel/SlideModes/SlideEditPanel/SlideContentEditor.js
+++ b/components/Deck/ContentPanel/SlideModes/SlideEditPanel/SlideContentEditor.js
@@ -1198,57 +1198,57 @@ class SlideContentEditor extends React.Component {
     placeCaretAtStart(id) {
         console.log('placeCaretAtStart');
         let el = $('#'+id)[0];
-//        let el = $('#'+id).find('span:visible:first').not('.cke_widget_wrapper')[0];
-//        console.log(el);
-//        if (!el) {
-//            el = $('#'+id).find('p:visible:first')[0];
-//            console.log('id + find first span not found');
-//            console.log('try id + find first p');
-//        }
-//        if (!el) {
-//            el = $('#'+id).find('div:visible:first').not('.ui-resizable-handle')[0];
-//            console.log('try id + find first div not ui-resizable');
-//        }
-//        if (!el) {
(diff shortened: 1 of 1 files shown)


kadevgraaf:
Hi Viktor!
How is the ID dulication solved (e.g., by making range a const)? As far as i can see/understand this is mostly about having the caret on the first element in input boxes.

vkovtun:
@kadevgraaf Hi Klaas! I just added a prefix to the ID. From what I could see, the element was queried by class name not by ID, I didn't see any changes in behavior. The only relevant line is this one:

    this.innerHTML = '<div tabIndex="-1" style="top: -32px; left: 0px; position: absolute; z-index: 90000000;"  class="context-menu-one ui button blue outline '+  $(this).attr('id')+'" id="context-menu-'+ $(this).attr('id')+'"><i tabIndex="-1" class="tasks icon"></i></div>' + this.innerHTML;

kadevgraaf:
Hi Viktor!
The control-enter behaviour is now improved.
You wrote 

>  just added a prefix to the ID. From what I could see, the element was queried by class name not by ID, I didn't see any changes in behavior. 

However, I now cannot delete elements via the context-menu anymore, so perhaps it has some effect.

vkovtun:
@kadevgraaf I've fixed that problem with deletion.

kadevgraaf:
Thanks. Delete works, but now the send to front/back and duplicate do not seem to work anymore :)

vkovtun:
I've made another fix. Please check.

abijames:
I am not exactly sure what this PR is meant to do and the context menu works as before. But I am still able to experience the following issues with input boxes:
* delete first character in input box and content appears in the context menu button
* if I select text in an input box overlapping another one, text in the input box behind can also be selected (see screenshot)
* if you duplicate an input box it has the same id (e.g. http://localhost:3000/deck/3935-1/slide/4748-3/4748-3:4)

image

vkovtun:
@abijames I could not reproduce any of the issues you mentioned. The text in the screenshot is selected in one input box. The selection background is translucent, therefore it appears as if the word "Row 1" is selected as well. I have tested duplication several times and could not see any duplicate ID's nor could I see them in the example you mentioned.

PR #817, opened 2018-03-28
SWIK-2100 removes N and H shortcut

huwf:
Removes support for these keyboard bindings because they conflict with screenreaders. Should be done with associated reveal.js PR


abijames:
@kadevgraaf  this is linked to https://github.com/slidewiki/reveal.js/pull/16 and I checked it with Huw when he was working on it

kadevgraaf:
Works well. Thanks!

PR #818, opened 2018-03-29
Change tooltip and aria-labels on slide controls

abijames:
Removed shortkeys from tooltips and simplified languages


kadevgraaf:
Works well in Chrome and FF. Thanks!

PR #821, opened 2018-04-02
Swik 2039 questions exam prototype

dpaun:
This is a prototype for the questions exam mode.


abijames:
@kadevgraaf we decided not to merge this PR as we are going to use a different approach.

kadevgraaf:
Okay. Did not know that. Thanks for letting me know. I'll close the PR then (it did look nice though 👍 ).

PR #825, opened 2018-04-03
Swik 2154 edit button padding

abijames:
Fix to reduce gap between edit button and content panel.

@kadevgraaf this is caused by the buttons in the right toolbar being made slightly larger when you combine 2 icons. This is a hack but should be fixed properly when we sort out icons. Can you check I haven't broken anything?


kadevgraaf:
Works well, Thanks!

PR #826, opened 2018-04-04
Swik 1945 user based deck recommendation

dpaun:
This functionality can be seen when opening 'Decks' from the dropdown menu and selecting 'Recommended Decks'.
The recommendations based on user likes (more info in the deliverable D3.5) are provided by the service currently deployed on a server in Pupin. In cases of cold start (e.g. the user has not liked anything yet) the recommended decks panel shows some popular decks (with highest number of likes and downloads).


dpaun:
Could we merge this PR soon, since the user based deck recommendation is a part of the deliverable which was submitted about a month ago?

abijames:
@dpaun one suggested change before this is merged. I would recommend changing "weight" to "Ranking" in the order option on the recommended decks page as that will be more familiar.

dpaun:
Thank you for the suggestion. I will change it.

kprist:
@dpaun can you please include somewhere (preferably JIRA) what this PR depends on, service-wise ? Currently this hardcodes a server over at PUPIN. We need to add the service in our infrastructure and also consider data migration issues (if the service uses any persistence layer). 

I don't think we should add this to master until then, at least not enabled. It would be a good middle-ground option to somehow merge it but disable this until the service part is done. That way we will avoid any merge conflicts we could possibly have if we delay merging of this PR.

A nice way to do that would need some further code changes: move the service configuration (e.g. `const recommendationServiceUri = 'http://slidewiki.imp.bg.ac.rs';` ) into the `microservices.template.js` file and make sure it's empty in our production deployments, effectively disabling the functionality. We can then keep it on experimental enabled to test etc.

PR #827, opened 2018-04-05
Swik 2103 support for nv3 d charts in front end

sednanref:


components/Deck/ContentPanel/util/ChartRender.js
```diff
@@ -0,0 +1,100 @@
...
+    static renderCharts(resize){
+
+        let charts = $('div[id^=chart]');
+
+        if (charts.length === 0 ) {
+            return 0;
+        }
+
+        for (let i = 0; i < charts.length; i++){
+
```
vkovtun: To make the code more readable, I think we should not have several empty lines in a row. They are also at the beginning of the block, which makes not much sense to me.

components/PresentorHTMLLayout.js
```diff
@@ -64,6 +64,8 @@ class PresentorHTMLLayout extends React.Component {
                 <script src="/custom_modules/custom-semantic-ui/dist/components/search.min.js" defer></script>
                 <script src="/custom_modules/custom-semantic-ui/dist/components/api.min.js" defer></script>
                 <script src="/custom_modules/custom-semantic-ui/dist/components/sidebar.min.js" defer></script>
+                <script src="/custom_modules/nvd3/d3.v3.min.js" defer></script>
+                <script src="/custom_modules/nvd3/nv.d3.min.js" defer></script>
```
kprist: First of all, these files don't seem to be included in the `custom_modules` directory. Second, as they appear to be unchanged (by us) libraries which I am sure have npm packages, they should be imported via npm, and not copied over to the `custom_modules`. Haven't tested anything yet however.

sednanref: Hi @kprist, I didn't see they weren't added in custom_modules, now they are, the thing is that they can be downloaded via NPM, however, they must be added as a script as explained here: https://www.npmjs.com/package/nvd3 , so I preferred to add these files in the custom modules folder. In any case, you can check it working on http://localhost:3000/deck/4135-1/a

kprist: We are using webpack so that we can use the npm packages (and also upgrade them via npm), and also include them in the page as in the directions. There are many ways to do that, for example this: https://stackoverflow.com/questions/31616235/using-d3-js-as-an-external-in-webpack.

But there are also more manual ways to do that, and that's what we're mostly using. Consider these lines of code:
https://github.com/slidewiki/slidewiki-platform/blob/e0e94f1ebb040c2baca6aac0d68e92deb65bc313/components/DefaultHTMLLayout.js#L26-L28

In combination with this: 
https://github.com/slidewiki/slidewiki-platform/blob/e0e94f1ebb040c2baca6aac0d68e92deb65bc313/server.js#L61

You can always ask around for tips on how to include client libraries. The `custom_modules` folder should be used only when we want to use something that we also need to patch, or don't have npm packages (currently only reveal, semantic-ui and some ckeditor plugins that don't even have an npm).

sednanref: @kprist I got it, I didn't know about that configuration on webpack! It's already done. :smiley:

kadevgraaf:
Hi @sednanref  ! I get [object] instead of the diagrams currently. Perhaps there needs a library to be included?

screen shot 2018-06-07 at 18 14 13

sednanref:
@kadevgraaf  there's no problem with libraries so far, but the issue was in the import-service, I already sent a PR: https://github.com/slidewiki/import-service/pull/41

sednanref:
hi @kadevgraaf ! I did a change that fixes the issue in presentation mode and also gets the right labels... the only thing I couldn't figure out is how to get the title from the import of the slide correctly... besides that, there's also a bug that doesn't resize the chart correctly when you access the slide with the chart directly, but it does it if you navigate until that slide... do you have any idea what could it be?

kadevgraaf:
Hi @sednanref 
Sounds good!
What do you mean with "the title from the import of the slide correctly" - the title of diagrams or of slides in general (the latter is a more generic issues that sometimes occurs with importing Keynote presentations or ODPs that have been converted to PPTX.

Also the _bug that doesn't resize the chart correctly when you access the slide with the chart directly, but it does it if you navigate until that slide_ might be generic - if you directly navigate to a slide edit URL there are also problems in loading. In this case it might be that a Nv3d related library is not yet  loaded on the inital page load...

sednanref:
@kadevgraaf oh sorry, I meant the title of the chart haha

kadevgraaf:
Works for me, although slide edit resize/drag/context menu is limited and chart title is missing - I think these can be fixed later.
I tested in Chrome and Firefox with deck/4384-1/test-charts/slide/36425-6/36425-6:11/view - by uploading got the tech.pptx
What do you think @abijames  ?

kadevgraaf:
Original: screen shot 2018-07-09 at 17 08 01

On SlideWiki: 
screen shot 2018-07-09 at 17 08 47

kadevgraaf:
We might also need to test with diagrams other than the pie charts in got-the-tech

kadevgraaf:
I tested again with more charts in powerpoint (see attached source pptx file - [charts diagram import nv3d.pptx](https://github.com/slidewiki/slidewiki-platform/files/2277670/charts.diagram.import.nv3d.pptx))
Result:
http://localhost:3000/deck/4501-1/test-diagrams/slide/37717-1/37717-1:10/view

The bar-chart on slide 7 looks okay, as well as the pie-chart in the got the tech PPTX (see above or http://localhost:3000/deck/4502-1/test-diagrams/slide/37728-1/37728-1:11/view), but many others are not yet supported (they only show [object Object])

sednanref:
@kadevgraaf I'll take a look at this

sednanref:
@kadevgraaf Hmmm indeed there's no support for many of them yet in the import service, I'll try my best to bring the needed support for all of them, but please take in consideration that there are "There are 73 distinct chart types supported by PowerPoint." as said here: [Some Python-PPTX reference](https://python-pptx.readthedocs.io/en/latest/dev/analysis/cht-chart-type.html) I will try also to refactor a little bit the code so the addition of further support for other chart types is easier in the import service, but this would need also some support in the platform, lot of work, but I'll try to support at least the chart types in that file you shared :smiley:

sednanref:
Hi @kadevgraaf , as I mentioned before, there is a huge variety of chart types in Power Point (73) and unfortunately, NVD3 has just 11, so after the small refactoring in the import service and some changes in the ChartRender in the front end I could give some support for the deck "charts-diagram-import" that you shared, but there are some important remarks about this, I will be a bit specific to be as clear as possible. 

That deck you shared contains 9 different chart types, (some are the same type but with some different configuration) so what I could do for everyone is the following (slide by slide):

ChartType in PPTX -> ChartType in SlideWiki with NVD3

1.- Bar3DChart -> displayed as a BarChart (2D)
2.- line3DChart -> lineChart (2D)
3.- ofPieChart -> PieChart
4.- Bar3DChart ( in form of spikes (?) ) -> BarChart (2D)
5.- area3DChart -> AreaChart (2D)
6.- scatterChart (with a line) -> ScatterChart
7.- BarChart (with Boxplots) -> BarChart
8.- surface3DChart (which looks very similar to a Bar3DChart but with other dimension) -> BarChart (in form of multiBarChart)
9.- DoughnutChart -> pieChart (with hole, so it looks like a DonutChart :9 ) 
10. RadarChart -> BarChart (there's no better equivalent in NVD3 ...)

You can check them here: http://localhost:3000/deck/4534-1/c01

Please let me know if you have any remark, recommendation or something that you would like to change :smile:

kadevgraaf:
Looks good. Thanks for the nice improvements and hard work!

PR #828, opened 2018-04-05
Added avatar style to user picture in collaborators list... otherwise…

sednanref:
… the height expands outside the div


kadevgraaf:
Thanks!
The problem might be resolved in the meantime, but the avatar is nice indeed :)

PR #831, opened 2018-04-10
SWIK-2138_close_btn_social_signup_accessible

TBoonX:
Made cancel btn a html button and no div.

In order to test this, add after line 355 in UserRegistration.js:
```
$(ReactDOM.findDOMNode(this.refs.modal_social.refs.wrappedElement.refs.SocialRegistration_Modal)).modal({
            closable  : false,
            onDeny    : function(){
                //nothing
                return true;
            }
        }).modal('show');
```
Or just merge it because its a tiny change.

In general the whole component should be redone somewhen (use the react modal and elements,  use this.context.intl.formatMessage instead of FormattedMessage)

### DIFF
--- a/components/User/UserRegistration/UserRegistrationSocial.js
+++ b/components/User/UserRegistration/UserRegistrationSocial.js
@@ -262,8 +262,12 @@ class UserRegistrationSocial extends React.Component {
         }
         return (
           <div>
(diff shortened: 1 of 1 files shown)


abijames:
@kadevgraaf tested and ok to merge

PR #835, opened 2018-04-13
SWIK 2134   Slidenamebug   decktree rename and slide edit mode results in error dialog

kadevgraaf:


abijames:
@kadevgraaf  looks good. Seems to have solved the bug as reported in SWIK-2134

PR #837, opened 2018-04-16
SWIK-1856 Record presentation rooms

rmeissn:
Needs to be tested with https://github.com/slidewiki/file-service/pull/15

I've implemented a video-recording feature, currently tested on Chrome and Firefox. It only works if a user is logged in and several HTML5/modern browser features are available. Haven't tested, but it shouldn't work in private mode tabs.

You are able to opt-in session recording which records slide changes and your voice (records locally only). If you want to the file-service to create a video, press the camera button (and have a look at the filesystem of the file-service).

Questions:
* Is the codestyle okay?
* Did I miss any try/catch or promise/catch statements?
* Is semantic-ui supporting something like "toasts"? Showing a modal to notify users of executions errors seems to be too annoying...
* Is the indexedDB and sessionStorage removed/deleted on page reloads/closes?


TBoonX:
Regarding UI:
I was confused that I am not able to start a recording after I saved one.
Also confusing was that I was not able to start a recording when I declined it at the beginning. Starting record at a later point is important.

rmeissn:
@TBoonX That's true. I defenitely recommend to investigate into those two functionalities. Nevertheless I would prefer to not implement this now/as of this PR, but in a seperate one.

PR #838, opened 2018-04-16
SWIK 2156 and SWIK 2155   initial zoom canvaspptx2html and No vertical scrollbar with non canvas content + no zooming

kadevgraaf:


abijames:
Works much better on initial load (tested on Chrome and Firefox) as the slide is now fitting in the window.

but I noticed that if you resize the browser (eg make wider) then the slide doesn't resize and the reset zoom always goes back to the initial size of the browser window.... I realise this may need to be another task and its not so urgent

PR #850, opened 2018-04-26
Adds Markdown editor

ali1k:
This is the frontend part for https://github.com/slidewiki/deck-service/pull/132
which allows users to enable markdown and author slides in markdown


kprist:
I merged the relevant deck-service branch for this, so it can be properly tested using experimental services, especially file-service and pdf-service etc. We need to check it markdown slides export correctly to thumbnails and pdfs and epubs etc.

TBoonX:
How do I manage to make a slides content interpreted as markdown?

ali1k:
@TBoonX you need to go to deck edit and enable the markdown authoring for the deck.

PR #851, opened 2018-04-26
Fix addition of user group I belong to a playlist

schatzopoulos:
Fixes issue about not being able to assign a user group that I belong to a playlist.


abijames:
Works as expected

@kprist  could you review and merge as Klaas is on holiday.

PR #858, opened 2018-05-07
fix the issue of loading the slide editor for the first time after page refresh

ali1k:
@kadevgraaf To solve the issue with loading of slide editor on page refresh, I applied the following trick:
- I pre-load the slide edit store when slide view store is loaded. It works as a workaround but we still need to solve the issue of making SlideEditor a stand-alone component so that people can directly go to the edit mode. In the current situation, if you don't view the slide, you can not edit it.


kadevgraaf:
Works well on my localhost. Many thanks!

PR #867, opened 2018-05-11
Swik 1474 create notification for comment reply

dpaun:
When someone replies to a comment, currently only the owner of the content (slide/deck) is notified. This small PR (and changes already merged on the activities and notification services) enables the system to notify the author of the comment.


dpaun:
Yes, I'll update the notifications test. Thanks for the suggestion!

PR #868, opened 2018-05-11
Fix tree node rename sending wrong selector data to APIs

kprist:


components/Deck/TreePanel/util/TreeUtil.js
```diff
@@ -1,15 +1,17 @@
 class TreeUtil {
     //build node URL based on the context
-    static makeNodeURL(selector, page, mode) {
+    static makeNodeURL(selector, page, mode, slug) {
+        if (!slug) slug = '_';
```
TBoonX: Why is this underscore needed in any case?

kprist: I added that so that a URL would always have a slug part and to make the the URL parser in fluxible-router less prone to errors wherein it might confuse URL parts. Just a bit of defensive code. Actually there was a bug when the slug was missing which I've since fixed, but I like to keep stabilising stuff around :)

TBoonX: OK. I do the same in the merged component/Util of TreeUtil and ComponentUtil.

components/Deck/TreePanel/TreeNode.js
```diff
@@ -135,13 +135,13 @@ class TreeNode extends React.Component {
...
         let nodeSelector = {
-            id: `${this.props.rootNode.id}/${this.props.rootNode.slug}`,
+            id: this.props.rootNode.id,
             stype: this.props.item.get('type'),
             sid: this.props.item.get('id'),
             spath: this.props.item.get('path')
         };
         let canEdit = !this.props.permissions.readOnly && this.props.permissions.edit && this.props.ContentStore.mode !== 'edit' && !this.props.showThumbnails;
-        let nodeURL = TreeUtil.makeNodeURL(nodeSelector, this.props.page, 'view');
+        let nodeURL = TreeUtil.makeNodeURL(nodeSelector, this.props.page, 'view', this.props.rootNode.slug);
```
TBoonX: This is the only code line where slug should be forwarded?

kprist: No, there are lots of places we still need to add it, but for now I just put it in a few and will see how we manage that later. I'm not sure it's the best idea to do it like this, and would like to revisit later. This PR is a bugfix, not a feature PR, let's focus on that :)

TBoonX: Thanks!

ali1k:
@kprist we have noticed an issue: https://slidewiki.atlassian.net/browse/SWIK-2231 which seems related to this PR. Can you check that as well?

kprist:
I don't think it is related, couldn't reproduce the bug you mention myself anyway.

kadevgraaf:
I did small tests (rename + rename via slide edit) and seems to work 👍

PR #869, opened 2018-05-16
Fix slide thumbnail links in subdeck view changing deck tree

kprist:


kadevgraaf:
Works well. I found a minor bug - https://slidewiki.atlassian.net/browse/SWAQ-804  - but this fix is worth it.

PR #871, opened 2018-05-18
Add French, Lithuanian in deck languages dropdown

kprist:


kprist:
@kadevgraaf this also includes some unrelated changes to intl/default.json we need to remember to run `npm run langs` when we change such strings.

TBoonX:
Please also update the AttachSearchForm and the SearchPanel, they have both the old language selection too.

kprist:
@TBoonX done!

PR #873, opened 2018-05-18
SWIK-2234 checking the new activity_type when adding it to the feed

dpaun:
This PR is a small bug fix for the ActivityFeedStore.
User's newly created activities are checked before added to the activity feed (e.g. 'view' and 'exam' activities shouldn't be shown).


kadevgraaf:
Works well. Thanks!

PR #880, opened 2018-05-31
SWIK-2086 mobile view of deck panel

pierorex:


abijames:
@pierorex I just tried to test this PR and branch swik-2086 but didn't get the expected changes in mobile view. Can you confirm that is the correct branch?

pierorex:
@abijames I have checked and confirmed it's the right branch. The problem might be that you're starting the mobile view mode of your browser after you already loaded the site. Could you try to start the mobile view mode and then reload the page to see if it detects the changes? Then you should be able to go to a deck and you'll be presented with something like this:


image

abijames:
@pierorex I have tried to look at this branch again but it needs to be updated and conflicts resolved due to changes in the deck service. Please can you have a look at this?

pierorex:
@abijames I just updated the branch with new commits from the master branch and these are the results in two different browsers emulating different mobile phones, in two different computers: one running Google Chrome and the other one running Safari.
screen shot 2018-07-20 at 16 03 04
screen shot 2018-07-20 at 16 03 56
screen shot 2018-07-20 at 16 04 18
screen shot 2018-07-20 at 16 04 54

abijames:
I have tested this as much as I can without using a mobile. I sometimes see some bugs but I think this is due to the browser replicating the mobile screen e.g. when I alter the browser window size I see the mobile menu in the header but the deck view does not change. 
image

I would suggest we merge this into master to allow for proper testing on mobile devices to see if the experience is consistent. @kadevgraaf are you ok with this?

kadevgraaf:
Yes 👍 
Thanks @pierorex  for the hard work!

PR #885, opened 2018-06-01
Simplify locale dynamic loading; also add flags in UI components

kprist:
This will make things easier for adding more UI languages in the future.


kprist:
Seems to fail on travis with 
`FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory`
works in my laptop of 16GB :) will need to investigate...

abijames:
@kadevgraaf this is ok to merger

PR #886, opened 2018-06-04
Swik 2229 fix user pictures in discussion panel

dpaun:
This is a very small fix for user images on discussion panel.
Currently, when a user image is empty, the default 'user-alt' image is shown.
These changes enable discussion panel to display 'identicon' user images instead.


kadevgraaf:
Works well. Thanks!

PR #887, opened 2018-06-04
Swik 1900 feature request add simple drawings to slide

sednanref:
Hi! Here is a new feature to create own simple images during slide edition, you can find it under the 'Other' options as 'Paint' (I don't know if it is the best location for it), any advice is more than welcome!


abijames:
@sednanref can you add aria-labels to the buttons in the drawing tool. These should match the tooltips. 

Also can you make the text next to Primary color, Boarder colour and Line/border width labels for the associated input fields (with HTMLfor and id so they are linked programmatically).

And can you change the spelling of color to colour with a u ..... EU uses the British English spelling....

sednanref:
@abijames sure! thanks for remark those points!

PR #888, opened 2018-06-06
Show recently used decks on the homepage if logged in

rmeissn:
This commit will hide the carousel and show recently edited decks instead, like on the My Decks page, if a user is logged into SlideWiki. This has been requested by some trial partners.


kadevgraaf:
@abijames needs to review this Before merging.

PR #892, opened 2018-06-06
Swik 1929 usergroups ui translation

TBoonX:
Exchanged strings with formMessage stuff in UserGroups and UserGroupEdit


kadevgraaf:
Looks good, thanks!

PR #897, opened 2018-06-12
SWIK-2236 activity store compares ids without revision part - small fix

dpaun:
This is a very small fix for the activity feed store. When finding the activity which should be removed, deck id without the revision part is used for comparison.


kadevgraaf:
Thanks for the fix!

PR #898, opened 2018-06-13
Included button in remove group member

jaumejordan:


abijames:
The button labels are fine @kadevgraaf this can be merged

PR #899, opened 2018-06-14
Swik 1287 translation ui decks and slides

TBoonX:
This is a major PR which adds the new translations model and workflow to the UI.
For the general translations model discussion see [Confluence](https://slidewiki.atlassian.net/wiki/spaces/SWIK/pages/220856321/Realizing+content+translation)

Whats changed:
* The Translation button now openes a new Modal (DeckTranslationModal)
* The new DeckTranslationModal shows the primary language and all existing translations of a deck plus (if being a signed in user and having the rights) provides a list of possible languages from which a new translation could be created (button to do this is also there)
* The right sidebar now shows the current language of the deck
* The UI distinguish between a decks primary language and one of the translations - Being on a translations enables the TranslationMode
* The Translation Mode deactivates buttons and fields and thus only renaming, content change, reordering is possbile
* Deck language lists are now contain the same elements (AddDeck, Search, ...) 
* Creating a new deck URL is now unified via components/common/Util (no TreeUtil or ContentUtil for this anymore)
* The current language is a query parameter of the URL
* Multiple adoptions to service changes
* Simplified deck revisions access
* Switching between languages reloads the page (could be improved later, but was too difficult atm)
* Using a newer version of Uglify for production build
* Using smaller lib for the slug stuff
* added lib for 5 character language code handling

This PR is about testing the design, workflow and detect early problems. In order to do this work on the Athena server under
[slidewiki.imis.athena-innovation.gr](http://slidewiki.imis.athena-innovation.gr/)
OR start this branch local while using the following service URLs:
http://slidewiki.imis.athena-innovation.gr/activitiesservice
http://slidewiki.imis.athena-innovation.gr/deckservice
http://slidewiki.imis.athena-innovation.gr/notificationservice
http://slidewiki.imis.athena-innovation.gr/searchservice
http://slidewiki.imis.athena-innovation.gr/fileservice


kadevgraaf:
Works very nicely. Cool feature/work!

I checked quickly on http://slidewiki.imis.athena-innovation.gr
Two minor issues are I think about timing/loading/updating in the user-interface:
Sometimes the language flag does not correspond to the written language in the info panel:
screen shot 2018-06-15 at 14 12 44

And sometimes the translation switch button does not immediately show the name of the original language when you want to switch at the slide level:
screen shot 2018-06-15 at 14 13 18
This does work after you click in the empty button in the dialog and I think after you do a hard refresh of the page.

The comments in the comment tab (and probably also questions/exams) are the same for both language versions, but I am not sure if it useful to split those out at a later stage.

kprist:
Update: since the service PR, https://github.com/slidewiki/deck-service/pull/136 has been merged, this PR can be reviewed as normal, using the services of experimental.

actions/translation/loadDeckTranslations.js
```diff
@@ -10,8 +10,9 @@ export default function loadDeckTranslation(context, payload, done) {
             context.executeAction(serviceUnavailable, payload, done);//TODO improve
             return;
         } else {
-            if (res && res[0] && res[0].language) {
-                res = res.reduce((a, c) => {a.push(c.language); return a;}, []);
+            if (res && res.length[0] && res[0].language) {
```
TBoonX: res && res.length > 0 && res[0].language
is correct, right?

components/Deck/InfoPanel/TranslationItem.js
```diff
@@ -0,0 +1,33 @@
...
+import {flagForLocale} from '../../../configs/locales';
+
+class TranslationItem extends React.Component {
+    render() {
+        // build the href!
+        let routeName = 'deck';
+        let navParams = Object.assign({ slug: this.props.slug || '_' }, this.props.selector);
+        let queryParams = { language: this.props.language };
+
+        let href = ['', routeName, navParams.id, navParams.slug, navParams.stype, navParams.sid, navParams.spath].join('/');
```
TBoonX: Why not using the Util component for URL building?

actions/translation/loadDeckTranslations.js
kprist: yes this is a mistake, but currently I don't use this action, don't know how that code made it in the commit :)

components/Deck/InfoPanel/TranslationItem.js
kprist: I started out using some other way to do this that would require less data than what I ended up using, and then forgot about the Util method. I'm not sure if it would work either, as it seems i couldn't access window.location from the component code, at least not the render() method, which kinda makes sense

TBoonX: Util.makeNodeURL(this.props.selector, 'deck', undefined, this.props.selector.slug, this.props.language) will do the job.
window.location is just used optionally at the beginning of the makeNodeURL method, at the end the URL is the same.

kadevgraaf:
Looks good! 👍 

A minor comment In the info panel "translation missing" is sometimes shown whilst I just saved a translation - this probably has to do with store updating. This is corrected after refreshing the page or navigating somewhere else and back. I think for this issue we can still merge and fix this later. 

I have not yet tested extensively (e.g. with comments/tabs/slide edit features / presentation mode)

kprist:
Hi @kadevgraaf, thanks for taking some time to check on this! We are still working on this with @abijames and @TBoonX there are some pending user experience issues plus some bugs, so it's not ready for merging yet anyway. It's a first version where some things may not be quite right, like comments, attach or search, in the sense that we have done very little in other services so that they know about multilingual content.

kprist:
Adding some bugs for this in case they're lost in slack: 
- with brand new decks the switcher does not include the "add new translation" so there's no way to start one
- in the language switcher, when you select a language that is in tree translations but not in node translations, it doesn't switch the selected language

and a suggestion:
- when you "add missing translation" for a slide, you get to slide edit. we should be consistent and redirect to deck edit as well so translators can edit right away

kprist:
Also, something is weird about the switcher with multilingual decks (mixed language) like deck id 4018: sometimes i see all two languages as options, sometimes not.

TBoonX:
I improved the UX and fixed some bugs.
This is the list of things I worked on:

* bug in the language switcher, when you select a language that is in treetranslations but not in node translations, it doesn't switch the selected language
* when you "add missing translation" for a slide, you get to slide edit. we should be consistent and redirect to deck edit as well so translators can edit right away
* Add translation (after language that is missing) should be a button
* Add tabindex
* Original language should not clickable if it is shown
* Hide node language if its the same as the deck


@kadevgraaf in the branch exissts an error with CKeditor. It seems not to be there at experimental:
When the editor gets opened the console says:
screenshot from 2018-07-09 18-11-17

kadevgraaf:
Hi @TBoonX !
Looks good / quite intuitive now.
I tested and it also works with markdown (see http://localhost:3000/deck/4408-1/test-nederlands?language=nl )

The bug does not appear on my local machine - it might have to do with changes to how we deal with submodules on your local machine - we lineheight plugin from submodules to napa:
  "napa": {
    "ckeditor-lineheight-plugin": "https://github.com/lotusgodkk/lineheight.git",
    "reveal.js-menu": "https://github.com/slidewiki/reveal.js-menu"
  },
The error message says: "plugin.js" so that makes sense 🤔 
Does it work if you remove all subdirs, except for /reveal.js from /custom_modules/ and do `npm i` ?

TBoonX:
If I do so, _npm i_ returns
```
Package cairo was not found in the pkg-config search path.
Perhaps you should add the directory containing `cairo.pc'
to the PKG_CONFIG_PATH environment variable
Package 'cairo', required by 'virtual:world', not found
gyp: Call to './util/has_lib.sh freetype' returned exit status 0 while in binding.gyp. while trying to load binding.gyp
```
I cleaned also the node_modules directory and did npm i.
Do I need to reload the submodules again?

PR #900, opened 2018-06-18
SWIK 2094 notify max 50 slides

kadevgraaf:


abijames:
The new text area needs to have id="attachSubdeckModalDescription"
Can you change text to below (and not in a span);
<p>Select a deck to attach from your My Decks list or search SlideWiki.,</p>
<p>We recommend that decks have a maximum of 50 slides per (sub)deck for optimum performance when viewing your presentation. If you wish to collate lots of decks then we recommend creating a playlist.</p>

kadevgraaf:
@abijames Thanks for the feedack. I have updated the text and changed the ID

abijames:
As discussed to improve accessibility:
* reinstate the off-screen message with class=sr-only and make the text in the message match the new one on screen
* make the id for aria-describedby match the ID for the off-screen message

kadevgraaf:
Thanks for the feedback! I pushed a new commit to try and fix this.

PR #905, opened 2018-06-20
SWIK 2285 Play Fork Translate toolbar buttons do not activate on Enter or Spacebar

kadevgraaf:


abijames:
@kadevgraaf works perfectly!

PR #911, opened 2018-06-25
Apply some minor fixes to work with new language codes in deck service

kprist:
This patch will make current master compatible with some language code fixes we have agreed upon: 
- we are going to be using two-character language codes ([ISO-639-1](https://en.wikipedia.org/wiki/ISO_639-1))
- a fix for the existing database data will be applied to the mongodb data using `truncatelangs` command of https://github.com/slidewiki/slidewiki-data-utils

This is a transitional code change that will be refined once the translations PRs are merged: https://github.com/slidewiki/deck-service/pull/136 and https://github.com/slidewiki/slidewiki-platform/pull/899.

This will also allow the deck service PR to be merged, so that we'll be able to properly review / test the platform PR.

### DIFF
--- a/components/AddDeck/AddDeck.js
+++ b/components/AddDeck/AddDeck.js
@@ -93,7 +93,7 @@ class AddDeck extends React.Component {
         else {
             wrongFields.title = false;
         }
-        if (language === null || language === undefined || language.length !== 5) {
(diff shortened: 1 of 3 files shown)


kadevgraaf:
Works well. Thanks!

PR #912, opened 2018-06-25
Remove hardcoded values for language,license in slide update

kprist:


kadevgraaf:
Did a quick check - looks good!

PR #913, opened 2018-06-27
Make Publication Status: All the default setting on "My Decks" page

kprist:


abijames:
All ok but one minor suggestion --- would be good to have "All" at the top of the publication status list as it is now the default

kprist:
@abijames I fixed ordering of options. Also changed ordering of recommended decks as iI think playlists should be always last in list

PR #917, opened 2018-06-28
Fix deck card buttons appearance after semantic ui upgrade

abijames:
to fix styling of buttons on the deck card component


abijames:
@kadevgraaf @kprist this is all good - can you merge?

PR #918, opened 2018-06-28
Swik 791 import preview

dpaun:
After importing a pptx, thumbnails of created slides are shown.


abijames:
Some comments:
* when I load it I don't get any buttons on the dialog to continue, just a cross on the top right to cancel the dialog. This takes you back to the import page. So a "Continue to deck" and "Cancel import" buttons would be good. Preferably above the slide thumbnails so that you don't have to scroll through a long list to get to them.
* from an accessibility perspective - all is good but you don't need the alt text on the images as they are repeating the same text. Rather than losing the alt-text (as it can help other users), you could add aria-hidden="true" to the images so the screen reader skips over them .

dpaun:
Sweet alert, used here, does not have the option to display buttons above content (where the thumbnails are shown). I have experimented with displaying thumbnails horizontally instead of vertically in order to deal with the problem of having to scroll through a list of thumbnails to get to the buttons. It looks ok to me, although the mouse wheel can not be used anymore for scrolling through the thumbnails.

kadevgraaf:
Looks good!
I added some text about loading preview images/thumbnails.
The horizontal scroll bar is not keyboard accessible, but I think we can fix this later.

abijames:
Can we add tabindex="0" to each thumbnail? This will then solve keyboard access.

Also I am concerned that the "publish deck" checkbox could be confusing (as it could be interpreted as completing import). Could you change the label to "Publish your deck for it to show in search results immediately"

Can you also change "View deck" to "Complete import"

kadevgraaf:
I'll fix those 👍

kadevgraaf:
Looks all good. Thanks @dpaun !

PR #921, opened 2018-06-29
SWIK 2315   reorder left slide edit panel based on graphical designer feedback

kadevgraaf:


abijames:
I still think adding content should be at the top of the menu with templates and properties below. 

We could also put "Add" before image, video and other to be consistent

PR #923, opened 2018-06-29
Swik 2048 fix license modal

Sole-Valero:
- now the Licence Modal uses Intl
- accesibility problems of the modal patched
- Url to terms and conditions added


kadevgraaf:
Thanks!

PR #924, opened 2018-06-29
Replace src with data-src to enable lazy loading

kadevgraaf:
I cannot see a lot of difference in performance, e.g., between the same
presentation on
http://localhost:3000/presentation/2201-1/test-a-big-big-big-deck/2201-1
/#/slide-12856-1 and on
https://platform.experimental.slidewiki.org/presentation/2201-1/test-a-b
ig-big-big-deck/2201-1/#/slide-12788-1 (without data-src)

### DIFF
--- a/components/Deck/Presentation/PresentationSlide.js
+++ b/components/Deck/Presentation/PresentationSlide.js
@@ -8,9 +8,18 @@ import SlideViewStore from '../../../stores/SlideViewStore';
 let sectionStyle = { 'top': 'unset !important'};
 
 class PresentationSlide extends React.Component {
+    constructor(props) {
+        super(props);
+        this.replaced = false;
+    }
     render(){
+        let slideContent = '';
+        if (this.replaced === false && this.props.content !== ''){
+            slideContent = this.props.content.replace(/ src=/g, ' data-src=');
+            this.replaced = true;
+        }
         return (
-            <section dangerouslySetInnerHTML={{__html:this.props.content}} id={this.props.id} />
+            <section dangerouslySetInnerHTML={{__html:slideContent}} id={this.props.id} />
         );
     }
 


rmeissn:
According to my tests it is a little faster for the initial load (e.g. useful for mobile users) and only preloads images for the first two slides. If an image has been once loaded, it is not discarded as the user navigates away from the slide. Thus all loaded images will stay in the DOM/Browser.
This changes requires the user to have an available internet connection all the time - if the internet connection drops the slideshow will lack images.

kadevgraaf:
Thanks for investigating this @rmeissn !!
How much faster is it?
Not requiring an internet connection is also important (although you can download/export HTML), so we need to decide whether the benefits outweigh the costs.
As a user I would rather much wait 1 second longer than risking that my presentation does not work anymore when internet goes down. E.g. at conferences I preload presentations in a separate tab a long time before I start presenting.
@ali1k  suggested having a switch to the presenter mode in which you can enable lazy loading, but this is extra work that again might not outweigh the benefits if the loading time is only 100 milliseconds less on average.
@abijames  what do you think?

abijames:
Regarding offline access - we have the HTML download for that which I have used at conferences. I think we should promote this as the ideal option where internet connection is unreliable.

kadevgraaf:
Thanks for the feedback!
@abijames  - What do you mean by "promote" ? should we have an option button for "normal" and "lazy" loading? or a tooltip "Lazy-loading speeds up when starting presentation of a  large deck. Use the offline version if you internet connection is slow or unreliability".

I think the switch button is better, as I would normally prefer non-lazy loading while presenting. We can then still have a tooltip.

rmeissn:
I will try to measure the speedup of this option, so we have some reliable data to decide if it's worth spending the time. @bwulff Could you setup all services at testing and the database of slidewiki.org? I would prefer to use a server that is not used much for these kinds of tests. I will probably be able to test on 03 to 05 July.

PR #927, opened 2018-06-29
Swik 1947 improve ranking for phrases

schatzopoulos:
* Fix snippet for search slide results 

Dependencies: 
* https://github.com/slidewiki/search-service/pull/22
* https://github.com/slidewiki/solrserver-config/pull/14


kprist:
@abijames @kadevgraaf if you want to review this, checkout http://slidewiki.imis.athena-innovation.gr/ it uses data dump from produciton, so for login if needed you should use credentials from there

PR #929, opened 2018-06-30
Übersetze die letzten 2 Absätze der Homepage

katrinleinweber:
Siehe Diff & https://slidewiki.org/?locale=de.


kadevgraaf:
Thanks for the translation!
I have updated the language strings accordingly in our Transifex translation system (http://www.transifex.com/slidewiki/) 
The translated text will be shown on slidewiki.org in about 2 weeks (after updating slidewiki.org with the new release).

kadevgraaf:
Hi Katrin,

No worries. Thanks again for the translation!
We normally download the translation results from Transifex and then indeed update via the Git workflow (all at once before we release once a month and update slidewiki.org).
Storing the translations in Transifex helps us keep an overview and update all languages at once, but your Git pull request was nevertheless valid/good translation :)

Having a contributing guideline is a very good idea. Thanks! I will make one now for translations.

Kind regards,

Klaas

PR #932, opened 2018-07-03
SWIK-2331 Fix slideshow links in Recommended decks panel

dpaun:
There were some changes in the DeckCard component which cause an issue with slideshow links in the RecommendedDecks panel. This very small fix adds slugs to recommended decks to fix this.


kadevgraaf:
Works well. Link to presentation mode works again! Thanks! 👍

PR #935, opened 2018-07-05
SWIK 1752 allow image backgrounds for slides

kadevgraaf:


abijames:
Can you put aria-hidden="true" on the background image and make it have a null alt tag ... alt=" "

Also, how can you delete a background image if you want to remove it?

PR #936, opened 2018-07-05
SWIK-2131 - Embed-UI-needs-to-have-a-Title-field

kadevgraaf:


abijames:
The UI indicates that I only need to add a title if I am using a URL to embed an item, it could be for embed codes as well. I tried to add a title alongside embed code but it didn't work.

kadevgraaf:
Ah in that case I can add it by checking whether title is empty in code and possibly replacing the title with content of the required form field

abijames:
@kadevgraaf this works well --- can be merged if updated to master

PR #939, opened 2018-07-10
SWIK-2275_users_display_name

TBoonX:
Used new user attribute in the components and services (step 1).

This task in ongoing - several service needs to be updated and merged.

Here is my list of needed TODOs regarding the whole task (which possible responsible devs):

1. Deck-service needs to add displayName to editors user object @kprist 
2. DeckPropertiesEditor in platform has to use the new attribute @kprist 
3. Handler.js in discussion-service has to use the attribute @dpaun 
4. Question-service or platform has to enrich data with attribute @stavmars 
5. ContentChangeItem in platform needs to use the attribute @stavmars 
6. Attach slide/deck in platform have to use the attribute @Sole-Valero 
7. Presentation rooms have to use the attribute @rmeissn 
8. Activity-stuff - small changes are needed on activities, discussions, notifications services @dpaun 
9. Search deck/slide components in platform code


TBoonX:
TODO: Add displayName to cookie and the user plugin

TBoonX:
I enhanced the cookie with the displayName attribute but on my machine the cookie does not contain it. I checked that the plugin gets the attribute but its just not set! I even cleaned the browser and my npm modules. Does it work on someone elses machine or does someone has an idea?

TBoonX:
This PR is ready for QA. We want to merge it even if the subtasks are not done yet.
During QA you should test the following:
1) User settings - try to save the a new displayName
2) User profile pages - check that displayName is used
3) UserGroupEdit - check that displayName is used
4) UserGroupDetailsModal - check that displayName is used

Sole-Valero:
And...what about the search deck feature? It is also used in attach deck.

schatzopoulos:
The display name is used 2 times in the my decks page under user photo. I think that one of them should be replaced either by username or by full name (or both of them)

snap

schatzopoulos:
@Sole-Valero Search user filter is updated in this PR https://github.com/slidewiki/slidewiki-platform/pull/946 it works with display names and indeed it is used in the attach deck/slide interfaces.

dpaun:
There was a bug in the notifications service. I think I've fixed it.
In order for notifications, activities and discussion to use display name, result of user-service POST /users should also contain display names for users. I believe that this is a very small change.

TBoonX:
@dpaun _displayName_ is an optional attribute because we introduced it just now. Thus its not given in users which havent set it yet. All the profile routes are returning addionally and optionally the attribute:
GET /user/{identifier}
GET /user/{id}/profile
POST /users

dpaun:
Yes, I see it now. Thanks!

TBoonX:
I tried everything out again and the user search field in the deck/slide search gets its users from the search api and not user api. Thus its more to do than just access an attribute.
Same for user search in attach deck/slide.
Is this what you meant in slack, @schatzopoulos ?

schatzopoulos:
@TBoonX yes, that is exactly what I was talking about. I have already made a PR about that here https://github.com/slidewiki/slidewiki-platform/pull/946 that I think is ready to be merged.

dpaun:
I can update the activity feed and notifications. I've already updated the services today.
Should I use this branch or create a separate one?

abijames:
@dpaun if you update this branch then we can test is all works together.

TBoonX:
Works well!

PR #944, opened 2018-07-12
Fixing a problem with import modal selected file display

dpaun:
There is a small bug introduced with the Import preview - when selecting a file, its name is not displayed in the import modal dialog.
This is a very small fix.


abijames:
IS there anyway to make the text in the modal below the heading left justified and shortened to:

Here is a preview of your slides. It may take a few seconds for the images to be created. You can use the tab key to move through the images.

kadevgraaf:
Hmmm, but now after confirming the preview modal nothing happens...
I get "Warning: executeAction for `addActivity` was called, but `SLIDES_PROGRESS` is currently being dispatched. This could mean there are cascading updates, which should be avoided. `addActivity` will only start after `SLIDES_PROGRESS` is complete."

kadevgraaf:
Hmmm, it might be due to my local config - 
"SyntaxError: expected expression, got '<'[Learn More]" 
Do you have the same @dpaun  and @abijames  ?

dpaun:
I think I've fixed it by adding a missing import. Strange that there were no error messages indicating that...

kadevgraaf:
Works again for me 👍 
@abijames  can you double-check?

abijames:
@kadevgraaf this is good to go!

PR #945, opened 2018-07-13
Using the selected theme for import preview thumbnails

dpaun:
Import preview currently does not use the selected theme when showing slide thumbnails.
This is a very small fix for this.


kadevgraaf:
Works well, thanks!

abijames:
works well for me too, thanks for quick turn around! @kadevgraaf  could you merge and then we can get it on the release branch

PR #946, opened 2018-07-19
Update user search dropdown in search form

schatzopoulos:
Updates the user filter dropdown in the search forms. 
Also works with display names as it is merged in user-service.


abijames:
Works well!

abijames:
@kprist  would you be happy to merge this?

PR #948, opened 2018-07-20
Swik 2314 add resulting image of paint modal as svg in slide

sednanref:
In this PR in conjunction with https://github.com/slidewiki/file-service/pull/19 made by @rmeissn we include support to upload SVG files to the file-service.

I also added some modifications to fix https://slidewiki.atlassian.net/browse/SWIK-2336 reported by @abijames , I think that with the new redimension of the file the input box size would fit perfectly now.


kadevgraaf:
Looks good and works well.
Tested in Firefox dev latest. I also tried combining the SVG with an uploaded PNG, and that seemed to work.
@sednanref  - How much work would it be to also allow for re-opening the image (and perhaps also other images) in the paint modal for editing?

sednanref:
@kadevgraaf I'm actually working on it now, I think it's very possible, I'm starting by trying to edit svg's uploaded to the file service... but I was trying to use a function provided by fabricjs to extract objects from a svg from URL but it seems that it is very problematic because of some CORS issues with the library... I'm trying to find a workaround for it :)

sednanref:
hi @kadevgraaf ! I added the edition of images and svg's , can you take a look at it? I would like some feedback from you and @abijames :D to edit and image or an svg just open the contextMenu of the element, and if it contains an image or an svg (that was already stored in the file service) it will import it to the paint modal and then you can do whatever the interface provides to you. In the case of images, they will be treated as a single object... it's also important to note, that when you 'edit' an image, the resulting edition will be added in the slide, but the original image will remain in the slide so the user will need to remove the original one manually, is this ok? In case of svg's there is no issue, the new SVG will completely replace the previous one.

kadevgraaf:
Looks very good!
Some remarks:

- Resizing does not work anymore - I think this is valuable to keep.
- The context menu edit button is nice! Is it possible to also detect whether an input box with image is selected to then automatically edit the image when you click on the paint modal in the left slide edit menu? I think this is what users would expect primarily (before they think to check the context menu buttons).
- The original image remaining is I think a minor issue, but should be fixed later on (as it will confuse users). I can also fix this quickly I think when I get back from holidays.

sednanref:
@kadevgraaf thanks for the feedback! Regarding your remarks:

* Yes, now that the resulting 'image' is a svg, the resizing doesn't work, for now, I will address this on https://slidewiki.atlassian.net/browse/SWIK-2007 as far as I have researched it is not so trivial, but there must be a concrete solution for this :)
* Hmmm I didn't think about that, so you recommend that in order to create a new image with the paint modal no image/svg should be focused when the user click the paint button in the left menu? 
* Yes... I was thinking of ways to do this, but it would take some time... the way to delete them is not so easy as edit a svg because the url of the picture is different when it is edited (because it becomes a svg).

sednanref:
@kadevgraaf I did a small change that seems to work for the resizing of the svg's, well... it's not actually a resize but a rescale, resizing only one variable (like height) it's not that trivial, but I think this could work for now, what do you think?

abijames:
I have tried to test this but have found a bug in the image upload that needs to be resolved before further testing (SWIK-2402)

abijames:
@sednanref SWIK-2402 fix is good thank you.
Resized boxes work well. I just noticed that when I tried to edit an image, it sometimes appeared as a large black box afterwards. I had more than one paint object in the slide and had been playing around so may be an edge case.

image

sednanref:
After talking with @abijames the issue with the black box couldn't be replicated... So I propose accepting this PR, of course, if there's no other observation. If for some reason the black-box bug reappears, let's open a new issue/bug in Jira to handle it in a different branch so that the improvements here could be merged.

kadevgraaf:
It works quite well I think. Thanks!
I was not able to reproduce the black image bug.
Nice that you can also edit existing images.

I am happy for this to be merged.
3 minor things we can address in separate tasks are:

- After editing an image, the original should be replaced with the newer version.
- The alt-tag and title of the original image could be copied to the image upload/add dialog (after paint modal) - otherwise you have to enter it again
- Editing existing SVG graphics in slides would be nice. This is I think due to some changes I made to prevent earlier problems with selecting and resizing SVG.

sednanref:
I am also very glad that you merged it :D I will comment further in the respective sub-tasks. Thanks for dividing them it's easier to tackle them in separate :+1:

PR #954, opened 2018-07-26
Swik 2389 details in user group list

TBoonX:
I added a view details button to the entries in the groups list which opens the groupsdetailsmodal.
I used existing stuff but had to update it partly.
WHile developing I encountered the "element with ID 112 not found" error which is very un-verbose thus I removed the usage of FormatMessage in components which are displayed when viewing the groups list. At the end it was an error in the Modal but if the reviewers agree we could merge this also.


rmeissn:
Tested it: Works quite well.
One remark: If I accidentally click on "Leave Group", I leave the group instantly. A modal asking "Really?" would be good, as buttons for leave group and group settings are quite near together.

components/Deck/ContentPanel/DeckModes/DeckEditPanel/GroupDetailsModal.js
```diff
@@ -51,31 +51,34 @@ class GroupDetailsModal extends React.Component {
...
-                </div>
-                <div className="fifteen wide column">
-                  <TextArea className="sr-only" id={'usernameIsALinkHint' + this.props.group.creator.userid} value={this.context.intl.formatMessage(this.messages.linkHint)} tabIndex ='-1'/>
-                  <a className="header" href={'/user/' + this.props.group.creator.username} target="_blank">{this.props.group.creator.displayName || this.props.group.creator.username}</a>
-                  <div className="description">
-                    {this.context.intl.formatMessage(this.messages.groupCreator)}
+        if (showContent)
+            members.push(
+              (
+                <div className="item" key={this.props.group.creator.userid}>
```
rmeissn: Would be better to use react-semantic-ui in the future. It's possible to mix semantic-ui and react-semantic ui.

components/Deck/ContentPanel/DeckModes/DeckEditPanel/GroupDetailsModal.js
```diff
@@ -51,31 +51,34 @@ class GroupDetailsModal extends React.Component {
...
+                  <div className="ui grid">
```
rmeissn: Images may be included as Floating images instead of using a Grid. This would ease the DOM - better browser efficiency

components/Header/Header.js
```diff
@@ -15,6 +15,53 @@ import updateTrap from '../../actions/loginModal/updateTrap';
 let MediaQuery = require ('react-responsive');
 class Header extends React.Component {
 
+    constructor(props){
```
rmeissn: I would prefer to revert these changes as it was way less code before and thus less to read/search for if someone needs to change something.

components/Header/Header.js
```diff
@@ -145,7 +192,8 @@ class Header extends React.Component {
 
 Header.contextTypes = {
     executeAction: PropTypes.func.isRequired,
-    getUser: PropTypes.func
+    getUser: PropTypes.func,
+    intl: React.PropTypes.object.isRequired
```
rmeissn: If you revert you don't need to import this context too - I guess less overhead.

rmeissn:
Addition: There is a popup for the Group Name input field stating "Group Name", that is already written in front of the input. Please remove the popup as it is unnecessary.

TBoonX:
I reverted the changed INTL statements.
@abijames is the Modal OK? Should the Group name popup be removed as Roy suggests?

TBoonX:
SWIK-2400 makes this PR obsolete.

PR #963, opened 2018-08-01
Swik 2327 change date on deck view to last modified

sednanref:


abijames:
@sednanref this looks good but can you change the label for the date to "Last Modified" to be consistent with the other screens. Also, we need to make sure all the translations are updated.

PR #964, opened 2018-08-01
Adding URL validation, Year validation and proper empty-title validat…

sednanref:
…ion.


components/Deck/ContentModulesPanel/DataSourcePanel/EditDataSource.js
```diff
@@ -7,10 +7,40 @@ import cancelEditDataSource from '../../../../actions/datasource/cancelEditDataS
 
 class EditDataSource extends React.Component {
     componentDidMount() {
+
+        $.fn.form.settings.rules.yearRule = (year) => {
+            let currDate = new Date();
+            let currYear = currDate.getYear() + 1900;;
```
kprist: Date.getYear is deprecated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getYear , please use Date.getFullYear()

components/Deck/ContentModulesPanel/DataSourcePanel/EditDataSource.js
```diff
@@ -94,7 +124,7 @@ class EditDataSource extends React.Component {
                     </div>
                     <div className="ui required field">
                         <label htmlFor="title">Title</label>
-                        <input type="text" ref="title" id="title" name="title" placeholder="Title" defaultValue={dataSource.title} autoFocus required />
```
kprist: Is this change just cleanup? Seems the title remains a required field even after this.

components/Deck/ContentModulesPanel/DataSourcePanel/EditDataSource.js
```diff
@@ -7,10 +7,40 @@ import cancelEditDataSource from '../../../../actions/datasource/cancelEditDataS
...
             fields: {
                 title: {
-                    identifier: 'title'
+                    identifier: 'title',
+                    rules: [{
+                        type: 'empty',
+                        prompt: 'This field cannot be empty.'
+                    }]
+                },
+                url: {
```
kprist: The URL rule seems a bit strict, as it requires a protocol such as `http` etc. Not sure about usability concerns, since people might casually type in the URL with the protocol, e.g. `www.google.com`. This rule currently prevents this. Most people are not aware of what a proper URL is, or that it's required to include the `https://` in front of it. @abijames or @ali1k could comment on this as well.

Ideally, the form should be smart enough to do the check and insert it automatically, same way as browsers typically do. It would be safe to insert `http://` in such a case (instead of https) as most sites will do the redirect to https, whereas the ones that don't support https will not be able to handle that.

kprist: In the original report, https://slidewiki.atlassian.net/browse/SWIK-1499 I also commented that the URL is not a required field in the case of Web Page / Web Documents. This is still the case, i.e. I can add a page with just a title and an emtpy URL and it will still be accepted. This needs to be done if this task is to be considered complete.

components/Deck/ContentModulesPanel/DataSourcePanel/EditDataSource.js
sednanref: Yes, simple cleanup, to show the error with the semantic-ui validation handler.

components/Deck/ContentModulesPanel/DataSourcePanel/EditDataSource.js
sednanref: Hmmm I get it, well actually the URL is not required, the only required field is the title, I didn't change that. Regarding the url validation.... hmmm yes, I didn't notice that the protocol was strictly required, but then are we going to assume that the url uses http or https always? Actually it makes some sense for me, but just want to clarify that.

kprist: As I said, a site that supports `https` normally handles `http` requests as well by redirecting them to `https`. Sites that don't do that are very few and should reconsider, as it's trivial. On the other hand, a site that does not support `https` understandably cannot handle it and redirect to `http` for example. I think browsers handle that on their own, i.e. if the site does not respond to `https` they try switching to `http` but that's not always the case. 

So the safer approach would be to add a `http` protocol *when the user does not define it*, if possible with some smart library/coding.

components/Deck/ContentModulesPanel/DataSourcePanel/EditDataSource.js
```diff
@@ -10,7 +10,8 @@ class EditDataSource extends React.Component {
 
         $.fn.form.settings.rules.yearRule = (year) => {
             let currDate = new Date();
-            let currYear = currDate.getYear() + 1900;;
+            let currYear = currDate.getFullYear();
+            console.log(currYear);
```
kprist: please also remove the console.log

sednanref:
hi @kprist what do you think about the new rule for url validation? it uses a regular expression to ensure that the url contains http or https protocol, if the user doesn't specify the protocol, it will add http by default to it.

kprist:
@sednanref it looks somewhat better than even what semantic ui uses for their rule: 
https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/behaviors/form.js#L1245

What i don't get though is the last part, after the word boundary:
`\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)?/gi`

I can see it's optional, what is this ? 

I would also suggest you don't use the `g` flag, not sure if you should. Finally, since you're using the `i` flag anyway, you probably could shorten `a-zA-Z` to just `a-z`/

sednanref:
@kprist you are right, with the /i flag the A-Z is redundant, and without the g one the validation should be better (maybe faster (?)) the last part of the expression if I'm not wrong is related to the path of the url (I forgot to add the '/' in the expression)

components/Deck/ContentModulesPanel/DataSourcePanel/EditDataSource.js
```diff
@@ -22,7 +22,7 @@ class EditDataSource extends React.Component {
             if (!url.startsWith('http://') && !url.startsWith('https://')) {
                 url = 'http://' + url;
             }
-            let expression = /https?:\/\/(www\.)?[-a-z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-z0-9@:%_\+.~#?&//=]*)?/i;
+            let expression = /https?:\/\/(www\.)?[-a-z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b(\/[-a-z0-9@:%_\+.~#?&//=]*)?/i;
```
kprist: why include `//` (twice) inside the last brackets `[]` part ?
and, I still don't get why we need the `\b` (word boundary) there

sednanref: @kprist yes, after checking with http://regex101.com  the '//' was a mistake, but the \b, for some reason I don't understand completely well, is the one that doesn't allow to enter urls like 'google.com123' or 'google.commmmmmmmmmmm'... even when the {2,6} is specified there.

kadevgraaf:
Works well!
Not making the URL compulsory is fine I think. Adding sources is voluntary anyway. To be complete we would also, e.g., need to make access data of a webpage compulsory, but forcing this will refrain people from using the sources tab as it becomes more effort to add sources.

PR #966, opened 2018-08-09
Swik 2343 show playlists in usage panel

schatzopoulos:
- Adds Playlists content module tab to manage playlists.
- Adds playlists in user menu


kprist:
This also applies to https://slidewiki.atlassian.net/browse/SWIK-2378 ( adding the playlist user menu item )

components/Deck/ContentModulesPanel/CollectionsPanel/CollectionsPanel.js
```diff
@@ -0,0 +1,171 @@
...
+                <div className="ui stackable grid">
+                    <div className="row">
+                        <div className="middle aligned eight wide column">
+                            <h3 className="ui header">{this.context.intl.formatMessage(this.messages.header)}</h3>
+                        </div>
+                    </div>
+                    {   (userId) &&
+                        <div className="row">
+                            <div className="eleven wide column">
+                                <Dropdown value={this.state.currentSelection} placeholder='Select to add a playlist' fluid search selection options={collectionDropdownOptions} onChange={this.addCollection.bind(this, selector.sid)} />
```
kadevgraaf: Making the onChange here an onClick or onEnter might solve the keyboard selection problem

components/Deck/ContentModulesPanel/CollectionsPanel/CollectionsPanel.js
```diff
@@ -0,0 +1,171 @@
...
+
+        // collections that the user has edit rights
+        this.userCollections = [];
+        if (this.props.DeckCollectionStore.collections !== undefined) {
+            this.userCollections = this.props.DeckCollectionStore.collections.documents;
+        }
+        
+        // find collection dropdown options
+        let deckCollectionIds = deckCollections.map( (col) => col._id);
+        let collectionDropdownOptions = this.userCollections.filter( (collection) => {
```
kadevgraaf: altering this .filter function might help to only show collections that the user is admin of, and can thus assign a deck to.

schatzopoulos: Actually, you can select from playlists that either you are the admin or you belong to a user group that is assigned to that playlist. So, I think that in both cases you should have the right to add decks to that playlist.

kadevgraaf: Ah ofcourse, I forgot about user groups :) 👍  . Would it then be also good to remove playlists from decks that you have user group access to? (I recall this is currently not possible)

components/Deck/ContentModulesPanel/CollectionsPanel/CollectionsPanel.js
schatzopoulos: I managed to solve that by using the semantic ui dropdown and not the react-semantic-ui one. Please check again

components/Deck/ContentModulesPanel/CollectionsPanel/CollectionsPanel.js
schatzopoulos: I am not sure what you mean by that. If you have edit rights to a deck you will also have the right to remove it from all playlists? no matter if you have edit rights on the playlist? Hmm, I am not sure if we want this.

But, if you mean that currently you cannot remove a deck from a playlist in the playlist's page despite the fact that you have edit right to that playlist, then you are right! I have included this enhancement in PR https://github.com/slidewiki/slidewiki-platform/pull/973

abijames:
Testing comments:
* I now have a Usage and Playlist tag - will we be removing the Usage tab as it is redundant ?
* I added a deck to a playlist which then appeared without a remove button (see screenshot). While this was for a playlist I did not own (although I was a member), I was able to remove the deck from other playlists where I was only a member.
image

On the playlist tab, could you:
* Add the following test below the Playlist heading: "This deck is part of the following Playlists:"
* Remove the created date on the playlist (I don't think this is needed in this UI)
* move the dropdown box & Add to playlist button below the list of the playlists
* Add a h4 heading above the dropdown list "Add deck to a Playlist"

Accessibility is all ok except for an issue with the dropdown that affects all components and need to be resolved on a different task

schatzopoulos:
@abijames I added the proposed text/heading and removed created date from this UI.
As the list with the playlists can get large, I think it is better to keep the dropdown & "add to playlist" button on the top of that panel.

abijames:
@schatzopoulos thanks for the changes, they're good.

@kprist re the bug with a playlist being added without a remove button, this happens when I try to add http://localhost:3000/playlist/15?sort=order ... once you've looked into this, I am happy for this to be merged.

kprist:
@abijames I've managed to recreate this, and it is strange indeed. This is an issue for me for any of the playlists: 15, 14, 17, same thing. What they have in common is the user group ( 20 ), and I added myself into that in order to recreate it. Once I changed the userGroup of the playlist 15 to something else, e.g. 14, it worked!

@schatzopoulos can you check the code to see where the problem may be ?

kprist:
@abijames @kadevgraaf does this PR look good now? seems you have requested some changes in the past

kadevgraaf:
Works well. Thanks!

PR #967, opened 2018-08-14
Swik 2253 html import

dpaun:
This PR enables HTML (reveal.js) import (including images) of a zip file containing previously downloaded SW deck. This gives us round-trip export-import. 
Note: Import of speaker notes is supported, but currently the pdf-service doesn't export speaker notes because of a typo in the handler.js.


kadevgraaf:
Works well! Thanks!

PR #968, opened 2018-08-14
Swik 2334 delete comment

dpaun:
This PR enables users to delete their comments. Deck owners are also permitted to delete comments.
If a comment has replies, it will not be deleted but instead its content will be hidden (in order not to loose its replies).


kadevgraaf:
Works well, thanks!
Would it be good to make/register an activity for deteling a comment?

PR #969, opened 2018-08-14
Swik 1278 create thumbnail preview of themes

pierorex:


kadevgraaf:
Hi @pierorex ! Thanks for the proposed preview.

It looks nice, but for usability/UX I think it might be better to have an explicit button in the pop-up carousel that says "select theme" (now you have to know that you have to click the theme (does not work in Firefox) and then click somewhere else to close the dialog)

Perhaps even better might be that the pop-up is directly shown as a component on the add-deck page (albeit a bit smaller, e.g., 100 pixels heigh and 350 pixels wide, which also shows the theme title - a bit similar to the carousel on https://glidejs.com/docs/ . This would replace both the dropdown and the button (which now seem redundant as they both say "select theme"), and in the end save several clicks for users.

Another alternative (which seems tricky to implement) would be to preview the thumbnail of a theme in a (non-clickable) pop-up/tooltip when you select it in the existing drop-down.

kadevgraaf:
Hi @pierorex ! 
Cool, thanks! 👍 
Sorry for the confusion due to the task description. I hope it is not too much extra work.

pierorex:
@kadevgraaf I found a bug that I cannot solve in the component: when I save changes to a deck's theme and then enter the edit page again, the theme being shown is the old one but, if I go back to the deck page and enter again to the deck edit mode, the theme shows the updated version.
 
The component is added to both the add-deck and edit-deck pages.

Can someone help me understand what's happening?

abijames:
Hi @pierorex - re the UI side of this, I will updating the deck edit and add deck UI in the next sprint as there are number of changes/additions required. Can we work on that together to get this component in correctly

pierorex:
Hi @abijames
Yes, of course we can work together. Do you mind telling me the ideas you had in mind? :)

PR #971, opened 2018-08-16
Swik 1993 show slide current resolution

vkovtun:


kadevgraaf:
Looks good. Thanks!
I made some minor improvements to the button text and the getting width/height.
One thing still missing is when the slide resolution changes when applying a template. Can you take a look at this @vkovtun  ?

kadevgraaf:
Looks good. Thanks!

PR #972, opened 2018-08-17
SWIK-2239 Slide view page for older slide revisions is broken

vkovtun:
Added null checks in the slide view panel, so when that page is opened for a previous version it does not fail on null reference.


kadevgraaf:
Works well. Thanks!

PR #973, opened 2018-08-20
Swik 2066 add decks to playlist

schatzopoulos:
Adds functionality to add decks in a playlist via a modal in playlists page similar to attach sub-deck modal. 
You need to have edit rights to the playlist and press "Edit" and "Add decks to playlist" buttons to see the modal.

Please provide feedback on usability and accessibility. 
Do NOT merge before Abi's approval


abijames:
Really nice,

I've pushed a fix to the branch to address a frustating accessibility issue with screen readers.

General comment:  as you can also remove decks from the playlist using the modal window, is it correct for the button to be labelled "add decks to playlist" and then "add deck" on the modal. Quick change could be to change the button on the modal to Save.

kadevgraaf:
Works nicely! Thanks!

PR #974, opened 2018-08-21
Swik 2405 paint modal after editing an image the original should be replaced

sednanref:


kadevgraaf:
For testing I added (in http://localhost:3000/deck/4449-2/_/slide/37334-3/37334-3:2/view) an uploaded screenshot of a slide, then edited it by adding SVG boxes in paint modal, but the original was not replaced. 
When I saved and reloaded the slide, again editing the pictures did not result in the original image getting removed after editing it.

sednanref:
hi @kadevgraaf  ! I tried to replicate the behaviour you described, but I couldn't could you give me the image, and the process step by step to try to solve this please? Thanks!

kadevgraaf:
Hi @sednanref !
Some more details about the steps are that I first used the 'normal' image upload (with image below) in the left slide edit panel, and, after uploading, I clicked "edit image" in the top-left context menu of the (containing div of the) image on the slide, and then in the paint modal I added some drawing stuff, and then I saved the image. The original image that I uploaded was still there, as well as the newly introduced image resulting from clicking the context-menu edit button. I did this again to the resulting SVG image (to test whether SVGs would be replaced after edit).
I attached the image:
screen shot 2018-08-21 at 15 12 05

sednanref:
@kadevgraaf I fixed the issue, I also merged already branch 2407 related to [SWIK-2407](https://slidewiki.atlassian.net/browse/SWIK-2407). Please check if both functionalities are working and let me know if everything is fine :smile:

kadevgraaf:
Hi @sednanref !
It works nicely for editing existing SVG images. These also get replaced when edited again.
For both SVG and SVG+uploaded images the replacement does not yet seem to work consistently. Sometimes when I edit an existing SVG image, the original image becomes really large (so something is happening to it :) but it is not deleted)
I tested in Firefox and Chrome on slides localhost:3000/deck/4507-1/_/slide/37857-1/37857-1:27/view and http://localhost:3000/deck/4550-1/_/slide/37856-2/37856-2:26/view

sednanref:
hi @kadevgraaf ! I got a bit confused by your last comment, but I found a bug with some 409 response from the file service, they are already solved. 

What do you mean by that it is not working consistently? Do you have one concrete example for the case of that size change that you mention? One inconsistency that I found is, for example, in slide 27 you have some horizontal bars, but they are horizontal because of some HTML style there, which is not part of the SVG standard, even when you save the SVG as it is there, the final saved "SVG" if you open it again is showed vertical, so when the paint modal is opened, it shows a vertical bar... but I think this is fine, because the edition that made it horizontal is out of the SVG, but an HTML styling... I don't know if I'm very clear about this, in the beginning, I got confused too. Also, what do you mean by "but it is not deleted"? The original SVG? Maybe this was fixed with my last commit. 

Also, in slide26 I cannot edit any SVG, no context box is being displayed, so I don't know exactly what to do, I checked with experimental and the same behaviour is happening, so it must be some kind of bug in master

abijames:
@kadevgraaf I am happy for this branch to be merged.

kadevgraaf:
👍

PR #976, opened 2018-08-22
Fix - showing 'Owned decks' instead of 'My decks' for other users

dpaun:
There was a small bug in my implementation of recommended decks functionality which caused 'Owned decks' to be displayed instead of 'My decks' (except on experimental). :( This is a fix for that.
I've changed back the condition for showing 'Owned decks' instead of 'My decks':
if(this.props.user.uname !== this.props.loggedinuser) {   ...

### DIFF
--- a/components/User/UserProfile/PrivatePublicUserProfile/UserMenu.js
+++ b/components/User/UserProfile/PrivatePublicUserProfile/UserMenu.js
@@ -39,25 +39,14 @@ class UserMenu extends React.Component {
         });
     }
     render() {
-        let deckRecommendationsMsg = this.context.intl.formatMessage(this.messages.recommendedDecks);
-        let deckRecommendationNavLink = (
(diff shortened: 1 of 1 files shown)


kprist:
Since in both stable and slidewiki.org, there are no recommendations, And on experimental, as you said, the bug does not appear, I am guessing I need to check with localhost to see the bug ? Currently I can't do that, neither on experimental, as it seems the recommendation service is unresponsive ?

dpaun:
This issue can be seen on slidewiki.org if you go to My decks (on the left side it is displayed Owner decks and Owner Playlists).
Our server hosting the recommender service has crashed due to some hardware problem. I guess I'll need to prepare a docker container for the service as soon as I can.

kprist:
Ah I see it now, missed it as I was looking at the title.

PR #980, opened 2018-08-23
SWIK-2423---Show-original-Iframe-content-in-editor

kadevgraaf:
I disabled the CKeditor menu (as described in https://slidewiki.atlassian.net/browse/SWIK-2423) and image overlay. This shows the Iframe content in edit mode, but disabled the possibility to edit the Iframe.
I think this disadvantage (no editing) is outweighed by the benefits of seeing Iframe content in the editor: The only important thing to edit would be the Iframe title and URL (width and height can be done by the resize handles in slide edit). Editing the Iframe title is important for accessibility, and the title is only detectable by users with screen readers I think, so authors that have accessibility in mind would have created a proper title when adding the iframe. An alternative is editing the title in the HTML source editor, or removing the Iframe and then adding the same Iframe with a different title. This is not much extra work. Editing the URL is, I think, in most of the cases the same idea as making a new Iframe (as a different webpage will be shown). Again this can be edited via the HTML source editor or by removing the old Iframe and adding a new iframe. 
I have made a task (https://slidewiki.atlassian.net/browse/SWIK-2424) to edit the title and URL of an Iframe via the context menu which then sends title, url, embed-code, and dimensions to the left slide edit panel, but I think it has low prio.


abijames:
@kadevgraaf works well! Happy for you to merge.

PR #981, opened 2018-08-24
SWIK-2429 fixing Publication year removed when edit Sources

dpaun:
This is a very small fix of a typo which prevented saving the year field when editing datasources.


kadevgraaf:
Thanks!

PR #982, opened 2018-09-03
Swik 2379 follow a deck

dpaun:
This PR contains improvements to enable users to follow decks and playlists. 
For users that follow decks (playlists), notifications will be created based on activities on decks (or in case of a playlist, activities on decks in a playlist).
Service side modifications (activities-service, notification-service) have already been merged.
The UI can be previewed in the following images.
image
image
image
image


dpaun:
I've corrected a small typo on the activities-service, so now a notification should be created when there is an edit (or some other activity) of a slide in a deck which user is following.

abijames:
@dpaun as discussed my feedback on accessibility plus changing name to subscribe:

On deckview and playlists
* change the button label to Subscribe
** "Subscribe"
** "Unsubscribe"
* need a different or additional means to show when deck/playlist is subscribed as we cannot use only colour.  For example if we group the icon with a circle 
<Icon.Group size='huge '>
    <Icon size='big' name='circle outline' />
    <Icon name='rss'/>
  </Icon.Group>
and then switch to a filled circle and inverted colour on the rss icon
<Icon.Group size='huge '>
    <Icon size='big' name='circle' />
    <Icon name='rss inverted'/>
  </Icon.Group>
)
image

On the modal on the notifications page:
* Change text of the modal dialog to "Manage decks and playlist subscriptions". This should be a H2 and have an id followingModalHeader
* change text in followingDescription to "Change your subscriptions. Activities for your subscribed decks and playlists appear in your notifications."
* sub-headings should now be H3 and text as "Subscribed decks" and "Subscribed playlists". 
* so deck/playlist titles with need to be H4
* change Unfollow buttons to the same x icon as used to remove decks from playlists. Add aria-label "unsubscribe from [deck-title]" - this label will needs to be populated from the deck or playlist title.

abijames:
Thanks Dejan. I am happy to this to be merged. @kprist or @ali1k please can you check the code and merge. Then we can get this feature in the release

PR #983, opened 2018-09-03
[SWIK-2432] adding print view

ali1k:
- appends sources and speaker notes to the print view
- adds one last slide for listing contributors and license information


abijames:
Hi @ali1k as discussed in the call, it would be good if the print preview centred the content (currently shown to the left of the browser windows). Also when I tested with deck http://localhost:3000/deck/229-2 I had some slides with 2 visible borders e.g. 
image

This may be a bug on the platform as there seems to also be 2 frames on the slide but it is less noticeable http://localhost:3000/deck/229-2/mooc-introduction-usability-accessibility-and-user-experience2/slide/1561-1/1561-1:7/view

ali1k:
I checked those slides and it seems there are some border-related css styles within those slides. I am not sure how to resolve it. With regards to centering slides in the print view, I will try to fix it. Would be good if someone who is good in CSS can help me though...

ali1k:
I think another change we would need is to change the pdfservice to use the print view as source for generating the PDF export of a deck. Maybe @athird can make that change?

ali1k:
@abijames I fixed the centering issue. Could you please check?

abijames:
@ali1k centred print view looks good. I am happy for this to be merged. I can't get PDF export to work at all the moment so maybe we should just remove it from the download options when we merge this.

PR #984, opened 2018-09-03
Swik 2349 allow background colors for canvas based slides

sednanref:
Added also functionality to remove background colours through the remove background button. Moved the "change background colour" to the "add other" section of the left edit panel.


sednanref:
Please note the last commit that I added to support background colours in presentation mode, at the beginning I tried with dom manipulation and so on, but it was getting deeper and deeper because of asynchronous rendering, then I thought to save the background colour in the slide collection of the deck-service... but then I thought of this small 2 lines solution that I tested in different decks and it is working so far, please let me know how you find it, we can take the storing background colour in MongoDB approach as well if you prefer.

abijames:
@sednanref works well thanks! I have made one small change to move the buttons to underneath the properties sub-menu as seems a better place for them.
 
@kprist could you review this code and if its ok then merge it?

ali1k:
I fixed one issue in the presenter mode where the split function was used without checking if it returns more than 1 components.

ali1k:
@sednanref I am having some issues when I test it in my browsers (Chrome and Safari) on Mac OS. In Safari, I do not even see the color picker. In Chrome I see the color picker but when I click on that, nothing happens!

abijames:
@sednanref on Windows, this works correctly on Chrome and Firefox but the colour picker does not appear on Edge.

sednanref:
@ali1k thanks for the feedback and the small fix! I will check this!

sednanref:
@ali1k @abijames what version of Safari / Edge are you using respectively? Does the colour picker work in the Paint Modal? Maybe the problem is that this input type is not yet supported in those versions... I work with Linux and both Chrome and Firefox are working nice for me... if it is about versions and support... maybe the best approach would be to create/use another colour picker. Please let me know.

abijames:
I have Microsoft Edge 42.17134.1.0 - in the paint modal the colour picker looks very different (hadn't noticed before)
image

ali1k:
I am using Google Chrome Version 68.0.3440.106 (Official Build) (64-bit) and Safari Version 11.1.2 (13605.3.8) on Mac. In the paint module on Chrome, I can still see and set the colors, however, it doesn't show up in Safari at all.
I see that now you are using a react-based color picker. That one seems to work on both Chrome and Safari. The only issue I have is that the OK and Cancel buttons on color picker do not work. Is there a way to remove them or make sure they will close the color picker windows?

sednanref:
@ali1k yes! since I think it is an issue on browser dependency I'm now changing the approach to a custom Colour Picker to avoid such dependency, but I just started experimenting a little with it yesterday, as soon as I'm done with it I will let you know, maybe I can have it ready for Thursday :)

sednanref:
@ali1k I think I fixed the functionality of the ok and cancel buttons in the colour picker, please let me know if it is working on your browsers now, it is working for me :)

ali1k:
@sednanref @abijames  I tested it in Safari and Chrome. It works well and I think is ready to be merged. Just one minor request: when you add a picture to slide, you can set it to be as background-image. We would like to have the same functionality as background color there as it now only renders the picture in the are of slide and not the whole screen. Do you think we can use some of your code to adapt that functionality as well?

sednanref:
@ali1k do you mean in presentation mode? I just saw what you mention, I wasn't aware... but yes... somehow I think we could not exactly reuse but to adapt a little bit more the code to add that functionality... but I would prefer to handle that in a different task/issue.

ali1k:
@sednanref yes, in the presentation mode. We will indeed create a new task for that. I will merge this one now to master.
@abijames I was just wondering if we want to move the background image also to the properties as it is somehow hidden now in the toolbars and people cannot easily notice they can use images as background for slides.

PR #985, opened 2018-09-03
Swik 1896 sweet alert on save deck edit is not read aloud

sednanref:
Hi @abijames , I had to add an OnOpen function in conjunction with jquery in order to this modal to work, for some reason, in this particular modal the AllowEnterKey is not working (even explicitly) or some other event that I couldn't find is blocking it. But after this change it seems to work, let me know if it's fine.


abijames:
@kadevgraaf this is fine to be merged

PR #986, opened 2018-09-06
SWIK2439 - Added handleKeyPress function

nk1g09:
Add the handleKeyPress function for opening the Paint Modal from the slide edit left panel.
Previously this function was referenced but did not exist in the code.


kprist:
@nk1g09 it would better to include this more detailed info you included here in the commit message instead, it makes it easier to review and merge. Also, just be direct in what you fix, no reason to add technical information, that should be in the code itself unless it's too complicated :)

PR #990, opened 2018-09-14
Lti group

urashidmir:
LTI user groups added.


kprist:
Managed to test the based instructions in https://slidewiki.atlassian.net/browse/SWIK-2534 by using a deployment in http://slidewiki.imsi.athenarc.gr . The result was a page in slidewiki platform with this: 
screenshot-1

The result seems to be that a user was created, and logged in. I was even logged in at the moment, as the user who created the LTI group, but i got logged out and logged in as this other user. This new user has the username, `admin@lti.org`, but the (fake) email registered was `lti1.user2@slidewiki.org`. The reason it's called admin I assume is because in the moodle demo, when I clicked open window, I was that moodles' admin. This new user was also added to the group i created in the My LTIs page.

First of all, we may need a test case where we actually embed a slidewiki deck in a moodle course. That's what I assume this feature is all about. Currently I don't know how to do that, maybe @abijames or @kadevgraaf can try it if they're more familiar with moodle or similar.

Other than that, it seems this still might need work, but I'm not sure. It seems the "My LTIs" pages were copy paste code from the user groups feature, which has since an updated UI. I say it's copy paste, because it's not broken and works like the "My Groups" worked before the recent merge. So this may be something for @TBoonX to look at as well and comment. 

Also, it seems that the user creating code might be creating duplicate usernames, which will probably break. @urashidmir is that possible ? This might also be related to the PR for the userservice, https://github.com/slidewiki/user-service/pull/58, that @TBoonX has wisely asked unit tests for, but agreed if things worked we might skip it. I'm not sure now that this "works", then again I don't know what this is supposed to do.

urashidmir:
I updated the test case. A Moodle user can specify the deck id in the LTI URL to embed a SlideWiki deck in Moodle. If no deck id is specified, "Welcome to LTI Login" page is embedded in Moodle.

The code won't create duplicate usernames. If there's a valid LTI request (i.e. valid URL and key) contains the username that already exists, the user will be signed in.

kprist:
@urashidmir consider this:

1. User A sets up an LTI to integrate SlideWiki content into moodle deployment A
2. User B sets up another LTI to integrate SlideWiki content into moodle deployment B
3. User C uses the course in Moodle deployment A. Their username in Moodle deployment A is "Umar". I suspect when they visit the course a user with username "Umar@lti.org" will be created.
4. User D uses the course in Moodle deployment B. Their username in Moodle deployment A is "Umar". 

What will happen when user D visits the embedded content ? Since the pair "url, key" will not be the same, I guess a new user will be created. What will be that user's username ? How can we test this in this or the PR for user-service ?

urashidmir:
<<<User C uses the course in Moodle deployment A. Their username in Moodle deployment A is "Umar". I suspect when they visit the course a user with username "Umar@lti.org" will be created.>>>

Correct

<<<User D uses the course in Moodle deployment B. Their username in Moodle deployment A is "Umar". 
What will happen when user D visits the embedded content ? Since the pair "url, key" will not be the same, I guess a new user will be created. What will be that user's username ? How can we test this in this or the PR for user-service ? >>>

If the User D has the same username "Umar" and valid pair "url, key" (even if different from the url-key pair of User C), the User D is signed is also as "Umar" and no new user will be created. The new user will be created only if there is a new username included in the LTI request.

kprist:
@urashidmir @athird is this behaviour desirable ? doesn't this mean two entirely unrelated people use the same Slidewiki account ?

athird:
We can't tell them apart in that scenario - the key-secret pair is intended to identify an LTI connection uniquely. This scenario can only happen if the user who set up the LTI connection has reused the exact same details in multiple configurations. We can warn them against it when they create the key and secret, but from the LTI Provider side, I don't think there's a way to tell where the request is coming from except by looking at the key-secret pair.

TBoonX:
> Other than that, it seems this still might need work, but I'm not sure. It seems the "My LTIs" pages were copy paste code from the user groups feature, which has since an updated UI. I say it's copy paste, because it's not broken and works like the "My Groups" worked before the recent merge. So this may be something for @TBoonX to look at as well and comment.

Depends on the requirements and needs for LTI. In general it would be user friendly (same UI and more features) to adopt the user group pages for LTI groups. But if shared decks, playlists and admin role is not important then keep it.

abijames:
Comments on UI for LTI settings:

1.  On /settings/integrations replace "users icon" next to My Learning Services  with "big graduation cap icon" . This should be black and not part of the link
2. On /ltis/overview please remove the icon "users" from the Add new service button and replace with plus icon. Also please alter text on button to correct capitalisation "Add new service"
3. On /ltis/edit please add a Cancel button on this page (colour secondary) and remove icon from Save button (and change text just to Save)

urashidmir:
> Comments on UI for LTI settings:
> 
>     1. On /settings/integrations replace "users icon" next to My Learning Services  with "big graduation cap icon" . This should be black and not part of the link
> 
>     2. On /ltis/overview please remove the icon "users" from the Add new service button and replace with plus icon. Also please alter text on button to correct capitalisation "Add new service"
> 
>     3. On /ltis/edit please add a Cancel button on this page (colour secondary) and remove icon from Save button (and change text just to Save)

Done

PR #992, opened 2018-09-17
SWIK-2316 horizontal thumbnail gallery

pierorex:


aoelen:
Sometimes I get an error when loading the deck view. I think this happens because the image gallery is using "slidesArr[0].id", while "slidesArr" can be empty because the slides are not loaded yet. 

The following message appears in the console: Uncaught Error: Cannot read property 'id' of undefined

PR #995, opened 2018-09-25
Swik 2468 allow background images to be correctlydisplayed in presentationmode

sednanref:
Hi @ali1k @abijames as discussed in the last PR about background colours, I modified a little bit the code used there and managed to adapt the background images to be used as the background of the full slide in presentation mode. I tested this in different slides/decks and using Chrome and Firefox in Linux, please let me know if this approach works for you :)


ali1k:
Hi @sednanref thanks, I tested it and it seems to work fine. @abijames shall we also add this one to the release?

abijames:
I see this error when trying to get presentation mode to load with a background image " error: uncaughtException: Cannot read property 'split' "

abijames:
@sednanref I have retested this. It works correctly when I create a new slide (on Chrome, FF, Edge) but just not when I added a background image to the particular existing slide I picked. So I am happy for this to be merged.

re @ali1k suggestion to move background images to properties, we could separate out the backgroun image upload by adding an "add background image" button and then replicate the Image import process there. This  alt-text input field can be disabled on this modal (background images are hidden from screen-reader)

sednanref:
@ali1k oh right! I'll check those split more carefully, I also like the suggestion from @abijames but I have a doubt, I'll comment it in Jira

kadevgraaf:
Looks good in Firefox and Chrome 👍

PR #996, opened 2018-09-25
hot-fix remove pdf option from download modal as broken

abijames:
Commented out option for PDF in the download modal as it is broken. New print view can be saved to PDF


bwulff:
Thanks for the quick-fix!

PR #997, opened 2018-09-25
fixes pdf export in download pdf

ali1k:


abijames:
@ali1k @bwulff this works well

@ben can you merge this into the r 2.10?

PR #1001, opened 2018-09-26
upgrade some npm packages

ali1k:


ali1k:
fixes the critical lodash issue "Prototype Pollution".

PR #1008, opened 2018-09-28
SWIK-2485 Fix that presentation mode won't start on a specific slide

rmeissn:
Bug discovered by @kprist in SWIK-862

### DIFF
--- a/configs/routes.js
+++ b/configs/routes.js
@@ -579,8 +579,8 @@ export default {
             async.series([
                 (callback) => {
                     // add missing sid in order to load the deck's title
-                    if(payload.params.sid === undefined)
-                        payload.params.sid = payload.params.id;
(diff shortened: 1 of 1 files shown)


rmeissn:
This PR works for decks with deck IDs <=488 and starts breaking for decks with deck IDs >=491.

kprist:
Reason it breaks: lower decks use also lower slide ids. There is bug here where it sends a request to /deck/{id}/slides but uses the slide id instead of the deck id. So it "works" when the slide id is also some deck id.
There was a deck service bug that returns 500 instead of 404 for missing deck ids, which is now fixed.

ali1k:
I tested it again and seem to work fine.
@bwulff @abijames I think we should also merge this PR to our r2.10.x release

PR #1013, opened 2018-09-28
Swik 2400 usergroup pages

TBoonX:
I created a groups page which has its own route/URL.
The groups overview is still the same but each group list item has a button to go to the groups page of each group.
The page looks similar to the user profile page and has the three sub-container: settings, decks and playlists. Everything except the group image upload does work.

@abijames please verify for now if the style is OK and I head in the right directions. There is fine tuning to do and the group image upload is not working yet.


rmeissn:
- I've changed the title and description of the group + clicked on save. Result: the unchanged values are displayed, but the changed values are displayed after a page refresh (f5)
- If I add a user (e.g. Klaas), this new user is listed as a member correctly. But it's not saved. I guess I need to click the "save button"?! This is not intuitive. Either save it directly by clicking on the user or add something like a "pending changes" area and add a "Do you want to leave this page without saving" dialog.
- "Remove Group Member" isn't working
- Group picture section changes my user profile picture
- clicking "my groups" sends me to a different page - I didn't expect it to do this
- "create new group" button on "my groups" results in 404 page

bildschirmfoto von 2018-10-10 11-26-00
Move these buttons to the right  + add spacing between them

Deleting a group seems to work.

kprist:
I just resolved some conflicts and checked this branch out. I can see this is quite an overloaded branch, as it tries to do too much at once. While @rmeissn  has located issued with the updating of the group details, there are still issues with the main feature I would to see merged, i.e. the title of the task: "Show decks and playlist that are managed by a user group".

If I am logged out or not a user group member and visit a group page, I can see:
* text areas I can type in the title and description (but not save, which is good). the name and description are already available for viewing to the left
* buttons next to each user for removing them (they don't work, that is also good, but I don't want to see them)

What's more, when I go to a group page, either logged in or not, I'd rather see the group page, *with the decks and playlist* first, rather to a screen where I can immediately change details and membership. Those should be available via some option after I visit the page, not as the default view of the page. Similar to a deck page, we don't show people the edit view as the default.

kprist:
@abijames could you please review this as well ?

TBoonX:
@kprist I changed that the group landing page is now the groups decks and the editable elements are now hidden when they could not be saved.
You are right, there is much in this PR and its more than the related task is about. I copied stuff from user profile components and realized its good to do everything in one PR. I should have split/add tasks and do one after another - my drive to create a rich group page overhauled my duties in taskmanagement - now you know xD

kprist:
OK, seems we're getting somewhere: 
+ when logged in, I still go to the details page by default ( /usergroup/8/settings )
+ that page shows just the members when not logged in -> chould change the header to "Members" instead of "Group details" 
+ in the menu to the left, show options: 
++ "Shared Decks" (default from any menu/link/button to the group page)
++ "Playlists"
++ "Members"
+ in a separate menu to the left, add a button only for admins/creator for "Edit Details" that points to the `/usergroup/:id/settings` page
+ the `/user/:user/groups/overview` page (groups overview) could be taken off the `/user/:user/settings/profile` altogether, so that when you click "My groups" it doesn't feel you get taken elsewhere, like @rmeissn pointed out. we already have a link to groups via the user top right menu.

TBoonX:
- fixed
- fixed
- fixed
- I changed the name of the third option depending on the users role in the group.
- I dont understand exactly what you mean but I would like to say:
We could remove the "My Groups" button in the group page but abi requested to have one there thus we could change the name to "Go to My Groups". Regarding  groups list we have a different way than github - a public user profile doesnt show the groups the user is involved with. I would like to keep it that way. I think the "My Groups" button in user settings should still be there.

kprist:
I like how you changed the third option to either "Settings" or "Members", looks good!

You're right, I confused myself there too :) What I'm actually suggesting is
1. we remove the "My Groups" button from the settings page, and leave it in the group page. 
2. when a user visits my groups from the top right menu, they only see the my groups link enabled (blue), WITHOUT the Personal Settings menu, plus the list:
image
3. when a user visits a group page, the Groups/My groups stays to the left side so that people can return to that page (Abi's request), but only if the user is a member of the group. 
4. Groups should be more discoverable, and that's is an overall goal, i.e. maybe we should show the groups for public users, but let's worry about that some other task.

Just two more things:
1. When I click "Group Details" I do go to the default, "Shared Decks", but the link there does not show active (blue)
2. Personal preference: the "leave group" button should be below the members list, and with a more prominent color (red for example)

abijames:
I agree with Kostits suggestion that we remove My Groups option from My Settings page but I don' think this is critical to hold up merging this branch. I would also like to improve the usability on the Groups listing page so that that it is more like the Playlist UI where the whole listing is a link - but likewise this is not critical for now.

I have pushed a commit to improve button labels etc to improve usability and accessibility. I am happy for this to be merged.

PR #1014, opened 2018-09-28
Swik 2479 wrong values in languageselector

TBoonX:
Fixes the missing deck primary language in the InfoPanel Dropdown when viewing a slide in a different language.
Also removes the direct to a deck when the slide translation is not existing when changing the language.


kprist:
Continuing work on this, as it needs more to work properly.. Also in combination with https://github.com/slidewiki/deck-service/tree/SWIK-2479_Fix_language_issues

kprist:
@abijames this should address the issues reported in the leipzig feedback, but since data on experimental is not the same as in production, I realise it's not easy to verify. I have verified it locally, but someone else should also be able to test this.

TBoonX:
I could test it. I could restore a sw.org dump and test everything locally. If thats OK, is the dump from 26.9 alright?

kprist:
Thanks @TBoonX I already had the same idea, so anyone can test this on http//slidewiki.imis.athena-innovation.gr/ .

@abijames the deployment there includes this PR, as well as https://github.com/slidewiki/slidewiki-platform/pull/1025, so you can test them both there.

PR #1015, opened 2018-09-28
SWIK-2475 Fixes mobile header bar issues

rmeissn:
This PR fixes that the header bar is displayed twice on an IPAD (emulation in chrome) and thus having two login modals opened up.
Furthermore I've increased the threshold to jump to the desktop view, as some buttons were not available on the homepage on e.g. an IPAD.

This won't fix the request to use em instead of px, as the framework has no support for em.

@abijames Is it okay to show the mobile bar till 1050 pixels? Before it was set to 768 pixels, but e.g. the login button wasn't displayed on screens that are e.g. 900px wide.

### DIFF
--- a/components/Header/Header.js
+++ b/components/Header/Header.js
@@ -91,7 +91,7 @@ class Header extends React.Component {
 
         return (
             <div>
-                <MediaQuery minWidth={768} values={{width: 1600}}>
+                <MediaQuery minWidth={1050} values={{width: 1600}}>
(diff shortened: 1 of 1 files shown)


TBoonX:
If em is not supported, I am fine with 1050 pixels.

rmeissn:
@abijames Do you propose to keep the search input in the header bar also on mobile devices/view? I'm currently not quite sure what you mean.

abijames:
@rmeissn this can wait for a separate task. Happy for this to be merged

PR #1020, opened 2018-10-02
Fix server timeout when visiting direct deck URL when logged out

kprist:


bwulff:
let's give it a try...

PR #1021, opened 2018-10-03
hot-fix remove pdf option from download modal as broken

kprist:


kprist:
Not needed anymore created by mistake

PR #1026, opened 2018-10-09
Swik 2441 show comments for all revisions

dpaun:
This PR enables displaying of comments for all (current and past) deck/slide revisions.
Service side (discussion-service) modifications have already been merged.


kprist:
Is it easy to make it so, if I'm on a deck with say **k** revisions, if I go to revision **n** page I only see comments on revision numbers **r ≤ n** ? Currently I see everything, (like you said, also in the future)

dpaun:
I agree that it would be better to show only comments from the current and previous revisions. I'll make the change.

kprist:
@abijames I've checked this seems ok, @dpaun seems to have implemented my idea, hope you like it as well :)

abijames:
I've checked this and happy for it to be merged

PR #1030, opened 2018-10-12
SWIK 1342 ContentRecomendationDeckPage

kadevgraaf:


kadevgraaf:
outdated reveal - to check

kprist:
Noticed @kadevgraaf comment about updated reveal.js, and from what I can see this branch is using current master of https://github.com/slidewiki/reveal.js.git whereas current master of platform uses an older commit, from 4 months ago. So what's the correct reveal.js library we want to use ? From reading here: https://stackoverflow.com/questions/43364526/git-submodule-is-not-shown-as-changed-after-merging-branch-where-it-was-updated I think we need to run `git submodule update --remote` while on master in order to get master to latest commit from the submodule, is that what we should do?

kprist:
After running `python app/reco/update.py` on https://github.com/slidewiki/recommender-service currently deployed on experimental, this branch can be tested as well.

@abijames I've removed the similar contents panel from the deck view page and added the similar decks in deck landing page, like in the mockup originally using the featured items.

@kadevgraaf once/if this gets merged, I assume master will get an updated slidewiki/reveal.js submodule pointing at latest commit. If that's not desirable, please speak.

As with other branches, once this is merged, @bwulff will need to merge https://github.com/slidewiki/slidewiki-deployment/pull/19 and deploy them to testing.

components/Deck/DeckLandingPage.js
```diff
@@ -237,6 +249,7 @@ class DeckLandingPage extends React.Component {
                                     </Segment>
                                     <Segment attached>
                                         <Header size="small" as="h3">Tags:</Header>
+                                        <div className="sr-only">These decks have been recommended as they are similar to this deck"</div>
```
kprist: There's a `"` out of place, and I'm not sure about the text here @abijames

abijames: @kprist fixed

kadevgraaf:
Works on my localhost :)

PR #1031, opened 2018-10-15
Swik 2245 fix dragging large images

aoelen:
Max-width property is added to newly inserted images


abijames:
This fixes the bug although the images are now imported at a very small size. @kadevgraaf if you are happy with this approach then I am ok for this to be merged.

aoelen:
The initial width of the images is set to 30% of the slide width. This can be changed to, for example, 50% if you think that is better.

abijames:
50% would be better.

Sent from my iPhone

On 17 Oct 2018, at 12:58, aoelen <notifications@github.com<mailto:notifications@github.com>> wrote:


The initial width of the images is set to 30% of the slide width. This can be changed to, for example, 50% if you think that is better.

aoelen:
Okay, just pushed the change

kadevgraaf:
Code looks good and works wel in FF and Chrome. Thanks! 👍

PR #1035, opened 2018-10-17
Swik 2383 support education levels

kprist:


kprist:
Deck service PR already merged: https://github.com/slidewiki/deck-service/pull/146

abijames:
I've push small corrections to the accessibility to the page. Note at the moment the Education level dropdown has an empty option and is set to be a required field in the code but this is not showing in the UI (probably semantic-ui-react bug) and also the form validation does not report that leaving this field empty is invalid. If this is ok for the service, I am ok for this to be merged.

kprist:
The field is NOT required by purpose. Are you OK with this? Is it otherwise OK ?

abijames:
required attribute removed on education level

PR #1036, opened 2018-10-17
Swik 2384 support topics for decks

kprist:
Issuing this PR so that people can review some ideas about the UI. Currently it's based on the tags UI (because topics have been implemented as special tags). The UI is only in Add Deck page for now, and it's similar to the edit tags UI in Tags Tab. Main difference is, it does not allow for user-submitted topics. 

For viewing topics, currently implemented is color-coding of topics mixed with other tags in tags panel. This is just to showcase that capability, if we like the idea we can fine tune the appearance, for example we could include an icon, show them first in line, etc. This is in contrast to the Add Deck UI where they are separate. We can mix and match or choose a consistent UI for editing and viewing.

This is dependant on the merging of https://github.com/slidewiki/tag-service/pull/5, as well as putting the codes included in http://uis.unesco.org/sites/default/files/documents/international-standard-classification-of-education-isced-2011-en.pdf (Annex III) as detailed in https://slidewiki.atlassian.net/browse/SWIK-2384. 

For the purpose of reviewing this on a deployment, please check out http://slidewiki.imsi.athenarc.gr .


abijames:
Pushed commits to improve layout and accessibility of Add decks page.

kprist:
@abijames  I've added support for editing topics in deck edit panel as well. Please check the UI, I've set it up to be similar to add decks page.

abijames:
Some issues:
* If there are more than 1 topic and they are removed via deckEdit, then they remain in the DeckView screen. Removing one at time works ok
* spacing between Description and Subjects on deckview is too small - I have pushed commit to fix this although spacing now large.

kprist:
I couldn't recreate the first issue at all. Not sure what you want me to do about the second.

PR #1040, opened 2018-10-18
Swik 2492 playlist description field bigger

aoelen:


kprist:
@abijames is this good to merge ?

PR #1053, opened 2018-10-23
Fixing the issue with not working edit mode for not canvas slides.

vkovtun:


kadevgraaf:
Hi @vkovtun !
Thanks for the quick fix.
However, non-canvas slide edit does not work when you try to type in slide edit - on enter you go to the speaker notes menu.
And I see you manually add 960*720 to non-canvas slides, which makes them kind of canvas based and we lose the reveal.js benefits.

        if (slideHTMLContent.indexOf(‘class=“pptx2html”‘) < 0) {
            slideHTMLContent = ‘<div class=“pptx2html” style=“width: 960px; height: 720px; position: relative; “>’
                    + slideHTMLContent + ‘</div>’;
        }

Instead can you try to look at measuring the innerHTML height (I recall there is already code for this, but I guess that does not work anymore).

Or perhaps even setting the height of the content div in slide edit to a hardcoded 1000 pixels would also work for now.
Even nicer would be if the window auto-increases size when you add content, I recall similar to how it was done in SlideWiki 1, because you then have a preview of how the non-canvas content would scale on a monitor/beamer in reveal.js.

kadevgraaf:
Fixed in https://github.com/slidewiki/slidewiki-platform/pull/1056

PR #1055, opened 2018-10-23
Swik 1440 improve slide links in notifications

dpaun:
Currently, slide links in notifications point to slide-view, which is not good. Using the recently added activity parameter, content_root_id, slide links (in newly created notifications, which contain content_root_id) will point to proper path within a deck (deck-view + slide).

### DIFF
--- a/components/Deck/ActivityFeedPanel/util/ActivityFeedUtil.js
+++ b/components/Deck/ActivityFeedPanel/util/ActivityFeedUtil.js
@@ -67,6 +67,8 @@ class ActivityFeedUtil {
         let msPerYear = msPerDay * 365;
 
         let elapsed = nowTime - thenTime;
+        if (elapsed < 0)
+            elapsed = 0;
 
         if (elapsed < msPerMinute) {
             return Math.round(elapsed/1000) + ' seconds ago';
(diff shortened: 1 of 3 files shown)


kprist:
Haven't tested it yet, but I can see you also include the spath, calculating it ? Very interesting and maybe we can use it for other things, however there are issues with this parameter in our URLs, in combination with including the revision for decks / slides. These can lead to 404 pages, for an ever-changing deck. If the deck revision is not the latest, then the URL will never break, but if we are not, then those things (the spath and the slide revision) can change. 

So what I am asking, can we simply not include the spath and the slide revision? You can keep the deck revision for now. Or are there other considerations for that? 

To be clear, we can store things in services however we want, I'm just talking about the URLs.

kprist:
You should be able to do that, in theory. Can you give me an example where removing the slide revision fails ?

dpaun:
Yes.
https://platform.experimental.slidewiki.org/deck/4676-4/slide/41138 fails
https://platform.experimental.slidewiki.org/deck/4676-4/slide/41138-2 works

kprist:
Hmm this looks like a bug / regression, let me check it

dpaun:
It can wait, I agree. 
I will apply it for subdecks in the next commit.

PR #1057, opened 2018-10-23
Swik 2504 download questions

nk1g09:
Download button in question content panel for selection modal & download of questions


abijames:
Tested, works well. This will need to have accessibility improved later.
@kadevgraaf can you merge?

kprist:
Download does not happen without any error or warning. @nk1g09 can you check this again?

nk1g09:
> Download does not happen without any error or warning. @nk1g09 can you check this again?

What browser is this in? I just checked it again and it works for me in chrome and edge.

kprist:
Firefox; I'll check again in Chrome and let you know, thanks!@

nk1g09:
> Firefox; I'll check again in Chrome and let you know, thanks!@

Ah yes, there was a bug in Firefox. Have pushed a change which should now work in firefox, edge and chrome.

PR #1058, opened 2018-10-23
change blue highlight/border to SlideWiki theme blue to meet contrast…

abijames:
… requirements


kadevgraaf:
This should work well. Thanks!

PR #1065, opened 2018-10-29
Revert "Swik 2103 support for nv3 d charts in front end"

kadevgraaf:
Reverts slidewiki/slidewiki-platform#827


kadevgraaf:
Revert because of IE11 problem identified by @aoelen 
This might be fixed by commenting/removing/updating a file.
Also revert for now because diagrams are not working in presentation mode.

PR #1068, opened 2018-10-29
Reintroduce nv3-d support

kadevgraaf:


abijames:
Do not merge! Currently break IE11

kadevgraaf:
To fix before merge:
code/library/include that breaks IE11 
https://slidewiki.atlassian.net/browse/SWIK-2528

PR #1076, opened 2018-10-30
Update intl assets from trasnifex; enable portuguese; disable catalan

kprist:
This updates translations from transifex. Also enables Portuguese (30% translated) and disables Catalan (3% translated).
Includes a fix for the locale flags missing everywhere, would like @sednanref to verify that as well, as it was his code that broke it originally.


kadevgraaf:
Works well for me.

PR #1078, opened 2018-10-30
SWIK-2543 Change visibility of questions download button

nk1g09:
Make Download questions button visible only if user has edit permissions.

Stops any user from being able to download the questions.


abijames:
@nk1g09 I tested this branch on http://localhost:3000/deck/4512-3/bank-of-example-questions/deck/4512-3 and while I no longer see the download button, I now see all the questions in the panel including those in the exam mode.

nk1g09:
@abijames How many questions are you seeing? This deck has a larger question bank with 11 questions, of which 5 are ticked as exam questions. 

When I view it logged in I see 11, if I sign out I can only see 6 questions.

abijames:
@nk1g09 sorry you are right --- all ok

PR #1080, opened 2018-10-31
Swik 2541 change redirect new deck

aoelen:
Fix redirects to deck view: after creating a new deck and for the history tab.


abijames:
All good!

PR #1081, opened 2018-10-31
removed the double pptx class added to slide in print view

ali1k:


ali1k:
It is good to include this minor bug fix also in the final release. It just prevent adding a double border to slides when it already has a border defined in the ''pptx2html" class.

PR #1082, opened 2018-10-31
Swik 2546 fix hidden publish checkbox 

aoelen:
Added semantic ui classes to checkbox for publishing imported deck


abijames:
All ok.

PR #1085, opened 2018-10-31
Changes banners in home page randomly

ali1k:
This will allow choosing any arbitrary images for the banner shown in the home page.
I have already added some example images in the `/assets/home/banners` folder but before merging this PR, we should add the correct, enhanced images there.


abijames:
EA and I have been looking at the images and we feel that they are not working well as the people and non SlideWiki content are very distracting. We feel the mockup with the laptop screen is much more appealing to people who do not know SlideWiki.

kadevgraaf:
Good point. Can you upload/link a high-res version of the laptop image?

ali1k:
> EA and I have been looking at the images and we feel that they are not working well as the people and non SlideWiki content are very distracting. We feel the mockup with the laptop screen is much more appealing to people who do not know SlideWiki.

@abijames yes, I also think the other images need to be improved as they include lots of distracting elements.And in the image with just a laptop screen, I don't think we can show much of the slidewiki whereas in others people and presentations are explicitly shown...

PR #1086, opened 2018-10-31
Swik 2549 fix play buttons for sub decks

aoelen:
- Fix for problem with playing subdecks for the "play presentation" button on top of the deck tree
- Fix playing nested subdecks for play button in content footer


abijames:
@ali1k @aoelen does this PR change the action of the play button on the decktree to play just a sub-deck? This breaks the model that the buttons at the top of the tree impact the whole deck which @kprist and I have worked to maintain (play, fork, language in the deck tree all apply to the whole deck tree).

PR #1088, opened 2018-11-01
Made content with no canvas have full height displayed.

vkovtun:


kadevgraaf:
Thanks for the fix!

vkovtun:
No problem. ;)

PR #1092, opened 2018-11-07
SWIK-2560 fixing questions store state reset

dpaun:
This is a very small bug-fix for the issue with the questions tab which was discovered by Luis, during testing.


nk1g09:
Although I have not seen the original bug this seems to work fine. I could not access the add question interface or select exam questions interface if I switched decks (either through search or through URL),

PR #1093, opened 2018-11-08
SWIK-1775 Multilingual presentation rooms

rmeissn:
This PR extends the presentations rooms feature by:
- participants can now choose to view slides in one of the available languages instead of the presenters choice
- participants are able to exchange the chosen language at any time (e.g. use the presenters choice again)

This feature works for all available translations of a deck, whether these are complete (100%) or not (partial translation).

In order to test it from localhost you need to run slidewiki-platform by `npm run dev` on this branch + 
1. go to some deck
2. open a presentation room as a tab
3. share the link (open the link in another tab to mimic a participant)

Use e.g. deck 4492 for testing this feature. This deck got 4 languages (full translation for en, es, de and partial translation for sundanese). Translation means that the title of a slide is different. I have not correctly translated the title by a dictionary! 

**IF** there appears a CORS error at the participant tab, you need to:
1. ssh into experimental
2. run `docker exec -it deckservice_deckservice_1 /bin/bash`
3. edit routes.js with an editor of your choice in order to add beneath https://github.com/slidewiki/deck-service/blob/0c94d54fb0cb2dc83bd145195b2a1ec0b61a86a1/application/routes.js#L996 this line `cors.origin.push('http://localhost*');`
4. save routes.js and restart the service by executing `docker restart deckservice_deckservice_1`


components/webrtc/presentationBroadcast.js
```diff
@@ -609,7 +615,7 @@ class presentationBroadcast extends React.Component {
             switch (data.cmd) {
                 case 'gotoslide':
                     if (!that.isInitiator)
-                        changeSlide(data.data);
+                        this.changeSlide(data.data);
```
TBoonX: If _this_ is available, then use it instead of _that_.

rmeissn: I'm not sure why I'm using this and that over the switch case statement. There might be a reason I can't remember right now. I wouldn't want to change it in this PR. I've added a TODO instead

rmeissn:
@TBoonX This seems to be an issue, indeed. I will investigate into the issue with this deck and update the PR.

components/webrtc/presentationBroadcast.js
rmeissn: Changing it to this seems to work fine, thus I'm using now this in the function instead of that.

rmeissn:
@TBoonX I fixed the issue for deck 4486

rmeissn:
@kprist @kadevgraaf So feel free to merge this feature whenever you think it is best ;)

kadevgraaf:
Thanks!

PR #1094, opened 2018-11-12
Swik 2577 add header and footer tags

aoelen:
Added header and footer tags for the site header and footer


abijames:
+1 - shows up in accessibility API

PR #1099, opened 2018-11-14
Swik 2593 fix moving buttons CKEditor

aoelen:
Fixes a bug where CKEditor buttons were moving when their values were changed


abijames:
Hi @aoelen happy with this but we have noticed that often CKeditor doesn't always load the first time to enter the slide editor. I experienced it again when testing this branch.

aoelen:
Hi @abijames , it looks like that the editor is sometimes not loading when the edit page itself is refreshed. Editing slides when coming from another page seems to be working fine. Can you confirm this? 

In this branch mostly css changes were made. So the bug is probably not caused by the changes made in this branch.

PR #1100, opened 2018-11-15
adding metadata to deck page layout

abijames:
Creating PR from Ali's branch. Add metadata to improve SEO


TBoonX:
I fixed code in the store but the problem with the code is that the route /thumbail/slide/... doesnt exist.
It should be a full URI of the fileservice: 
`https://fileservice.experimental.slidewiki.org/GET /thumbnail/slide/{id}/{theme}`

kadevgraaf:
Thanks!

PR #1102, opened 2018-11-15
Swik 2603 question download button visibility

nk1g09:
Simple fix to hide the download questions button (within content panel) if there are no questions on a deck.


bwulff:
looking good

PR #1103, opened 2018-11-16
Improves headings, regions, UI consistency etc

abijames:
For SWIK 2514 - improves heading hierarchy and regions
SWIK-2595 - consistent colour of Unlisted label and remove from focus order
SWIK-2561 - edit message on ContentActionHeader re small screens instead of mobile

### DIFF
--- a/components/Deck/ContentModulesPanel/ContributorsPanel/ContributorItem.js
+++ b/components/Deck/ContentModulesPanel/ContributorsPanel/ContributorItem.js
@@ -7,10 +7,10 @@ class ContributorItem extends React.Component {
     render() {
         return (
             <a className="item" href={'/user/' + this.props.data.username} target='_blank'>
-                    <div className="ui image">
+                    <div className="ui image" >
                         <UserPicture picture={ this.props.data.picture }
                             username={ this.props.data.username } link={ false }
-                            private={ false } size={ 'mini' } avatar={ true } width= { 24 } bordered={ false } alt={'""'} aria-hidden="true"/>
+                            private={ false } size={ 'mini' } avatar={ true } width= { 24 } bordered={ false } alt={'""'} />
                     </div>
                     <div className="content">
                         <div className="description">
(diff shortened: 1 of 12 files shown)


components/Deck/ContentModulesPanel/ContributorsPanel/ContributorItem.js
```diff
@@ -10,7 +10,7 @@ class ContributorItem extends React.Component {
                     <div className="ui image">
                         <UserPicture picture={ this.props.data.picture }
                             username={ this.props.data.username } link={ false }
-                            private={ false } size={ 'mini' } avatar={ true } width= { 24 } bordered={ false } alt={'""'}/>
+                            private={ false } size={ 'mini' } avatar={ true } width= { 24 } bordered={ false } alt={'""'} aria-hidden="true"/>
```
kprist: I don't think this has the desired effect, from what I can tell by inspecting the resulting html. If we want this for all `UserPicture` instances, this change should be included in `UserPicture` file.

components/Deck/InfoPanel/PresentationsPanel.js
```diff
@@ -159,16 +159,16 @@ class PresentationsPanel extends React.Component {
         else
             toReturn = <div ref="presentationPanel">
                 <div className="ui basic segment" style={panelDIVStyles}>
+                    <h4 className="ui  header">Live Sessions</h4>
                     <Grid columns={2}>
                       <Grid.Column width={4}>
-                        <Popup trigger={<Button icon onClick={this.openChooseASessionModal.bind(this)} aria-label="Create a live session, Beta"><Icon name="record" size="large"/></Button>} content='Create a live session (beta)' />
+                        <Popup trigger={<Button icon onClick={this.openChooseASessionModal.bind(this)} aria-label="Create a live session"><Icon name="record" size="large"/></Button>} content='Create a live session (beta)' />
```
kprist: why remove the beta from the aria label only, and not the popup text?

components/Deck/ContentModulesPanel/ContributorsPanel/ContributorItem.js
abijames: moved aria-hidden to Userpicture component and remove role=presentation (as presentation role not as well supported by screen readers)

components/Deck/InfoPanel/PresentationsPanel.js
abijames: removed beta from tooltip

PR #1104, opened 2018-11-19
SWIK-2360 - Add check to hide speaker notes elements in slide view with content '' and ' '

kadevgraaf:


abijames:
@kadevgraaf works well. Happy for you to merge.

PR #1120, opened 2018-11-29
2248 no unsaved changes warning when navigating

vkovtun:


components/Deck/ContentPanel/SlideModes/SlideControl.js
```diff
@@ -16,8 +18,32 @@ class SlideControl extends React.Component {
     }
     componentDidUpdate(){
         this.updateProgressbar();
-        //ReactDOM.findDOMNode(this.refs.slideControl).focus();
     }
+
+    confirmLeaving = () => {
+        const hasChanges = this.hasChanges();
+        if (this.props.ContentStore.mode === 'edit' && hasChanges) {
+            // eslint-disable-next-line no-alert
```
vkovtun: We can rewrite the code to use custom dialogs instead of the browser's ones, but it is not very obvious. Moreover, when the user is trying to navigate away by changing the URL or closing the tab etc. they are presented with the browser's dialog anyway.

kadevgraaf:
Works well for me in FF dev and Chrome. Thanks for the fix @vkovtun !
@abijames or @kprist  do you want to doublecheck?

PR #1121, opened 2018-11-30
SWIK-2585 Open exam mode in a new tab

dpaun:
As requested, the Exam mode is opened in a new tab.


abijames:
Hi @dpaun the branch works well if you use the Exam mode button. However, if you embed a link to the exam in the text of a slide (as done in http://localhost:3000/deck/5026-1/_/slide/42328-2/42328-2:6/view) then the Cancel button doesn't seem to do anything. I am looking at this use case as my colleague decided this was the best way to use exams in a deck.

abijames:
All works well - thanks!

PR #1122, opened 2018-11-30
Swik 2431 add a source to a deck

dpaun:


dpaun:
This PR enables datasources on a deck level. For it to work, the corresponding deck-service branch needs to be used.

kprist:
@dpaun @abijames what's the policy on sources wrt to other than latest deck revisions? In terms of content, we don't allow any editing, whereas it seems we didn't apply this to sources. we also haven't applied this to tags either...

also, is this feature part of the functionality described in the final features deliverable ? asking for priorities.

abijames:
@kprist this was a feature request from trials and also allows users to handle the situation where they are loading up decks for another author (the then need add a source to the whole deck).

Re revisions - I say keep it as simple as possible. Happy for there to be no editing.

kprist:
@abijames not sure what you mean by "simple as possible". If you mean less code/changes, right now editing sources (deck or slide) are allowed even for older revisions of decks (they should be read-only), so minimal change would be to merge this almost intact (as well as the deck service PR). 

If you mean we should simplify edit rights so that other than latest deck revision should be uniformly read-only (including tags and sources), we need more work on this.

abijames:
@kprist i meant minimal implementation

PR #1123, opened 2018-12-03
SWIK-2610 Initial fix for incorrectly loading deck edit

dpaun:
This PR disables the 'Edit' button (when clicking on a decktree node, breadcrumbs, navigation buttons) until content loading has been finished.

### DIFF
--- a/actions/decktree/updateMode.js
+++ b/actions/decktree/updateMode.js
@@ -1,0 +1,7 @@
+const log = require('../log/clog');
+
+export default function updateMode(context, payload, done) {
+    log.info(context);
+    context.dispatch('UPDATE_MODE', payload);
+    done();
+}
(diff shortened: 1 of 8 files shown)


dpaun:
Thanks, @kprist . It is in the loadDeck now.
The mode is changed when "UPDATE_CONTENT" is dispatched.

kprist:
OK, looks better now. Can I ask, do we really need this `updateMode` action ? Couldn't we have directly issued a `context.dispatch('UPDATE_MODE', {mode: 'loading'})` command ? I can see we do various dispatches during other actions. Asking mostly because of some related code I'm writing as well.

dpaun:
Right! I'll change it. (It was a leftover from when it was triggered by a click.)

PR #1124, opened 2018-12-03
SWIK-906 Robots.txt

rmeissn:
Added plugin for serving robots.txt and added a robots.txt file.
I don't know about the content of a robots.txt file. I guess the current content means: Do not look at anything under this domain - this needs to be changed for at least slidewiki.org!


kadevgraaf:
Thanks!

kadevgraaf:
@bwulff  - can you check if this is suitable for slidewiki.org? 
I suppose we want robots to index most of our stuff except for what @TBoonX  disallowed.

PR #1125, opened 2018-12-05
SWIK 2612 single slide content translation modal

kadevgraaf:


abijames:
Works ok but the option is quite hidden.

Suggested changes to the text on the modal in the attached screenshot
image

PR #1126, opened 2018-12-07
Swik 2616 advanced options on search screen not accessible with keyboard

abijames:
On search page and info panel


abijames:
@rusnewman code. Checked for accessibility and usability 
@kprist @kadevgraaf please can you check code and merge if ok

PR #1128, opened 2018-12-11
Swik 555 fix deck not found handling page

kprist:
This mainly fixes the deck not found page for landing and deck view pages.

Related to [SWIK-1737](https://slidewiki.atlassian.net/browse/SWIK-1737) as it fixes one part of it: show an error message, but does not include the part about informing users the deck was removed.

Also fixes [SWIK-2526](https://slidewiki.atlassian.net/browse/SWIK-2526), as well as improving some loading indicators when switching between nodes in deck view.

### DIFF
--- a/actions/loadDeck.js
+++ b/actions/loadDeck.js
@@ -40,7 +40,7 @@ export default function loadDeck(context, payload, done) {
 
     // resets the deck view store
     // TODO (what other store to reset ???)
-    context.dispatch('LOAD_DECK_CONTENT_START');
+    context.dispatch('LOAD_DECK_PAGE_START');
 
     if (!(AllowedPattern.DECK_ID.test(payload.params.id))) {
         context.executeAction(deckIdTypeError, payload, done);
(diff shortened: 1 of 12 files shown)


abijames:
@kprist when testing this branch with eperiemental's services I had the following behaviours, please confirm if as expected:
* on http://localhost:3000/deck/1 I got a loading indicator that continued but with no messages
* on http://localhost:3000/deck/43311 I got a 404 error

As there are no suspended users on experimental I can't see the impact on their decks

kprist:
@abijames, the 404 error is the intended result for unknown deck ids or decks no longer available. For suspended users decks will also show a generic 404 page without any more information. 

About the `/deck/1` issue, it's a matter of improper data perhaps, as the deck does exist, but lists user id: 1 as its owner, whereas such a user does not exist on experimental. Since this is unexpected we could return an error page. For now I have patched the issue by omitting any user information.

abijames:
@kprist ok happy for you to merge

PR #1129, opened 2018-12-12
Add Open Graph meta data

aoelen:
Some Open Graph meta tags are added to the deck page. This is useful when SlideWiki links are shared via social media. Most social media sites will preview the title, thumbnail and description of a deck.


kadevgraaf:
Pushed to master and experimental to test whether this works (social media cannot access pages on my localhost).

PR #1130, opened 2018-12-12
Swik 2579 fix loading sub deck theme

aoelen:
On each slide changed event, the slide theme is set as main theme


kadevgraaf:
Works well in Firefox and Chrome. 
I get a flattenchildren warning:

_Warning: flattenChildren(...): Encountered two children with the same key, `31318-1`. Child keys must be unique; when two children share a key, only the first child will be used._

But even with the warning I am happy for this to be merged.

aoelen:
The flattenChildren warning was caused by something else (which is not related to this branch). So if that was the only problem, this branch can be merged. 

(A PR for a fix of the flattenChildren warning can be found here: https://github.com/slidewiki/slidewiki-platform/pull/1139)

PR #1131, opened 2018-12-12
CKEditor update button focus color and skin building

aoelen:
Update focus and hover color for CKEditor buttons (for better accessibility)


abijames:
works thanks!

PR #1132, opened 2018-12-12
Semantic annotations

aoelen:
Semantic annotations for slides. Can be tested from the slide editor. 

For automatic annotations: Edit slide > Add > Add other > Auto annotate 

Adding or deleting annotations manually: Right click on selected text/annotation, then add, edit or delete annotation.


kadevgraaf:
Works well in Firefox and Chrome. Thanks!

PR #1134, opened 2018-12-13
Swik 2615 tags are removed when selecting a file on the add deck page

dpaun:
This should fix the removal of tags when selecting a file for import or clicking on Add Deck button (while there are errors on the Add Deck page, e.g. checkboxes are not selected).
Tags are saved in the Import store, and this is used as the initialTags for the TagInput component.


abijames:
All works well thanks!

PR #1136, opened 2018-12-13
Swik 2605 skip to content test

nk1g09:
Link that appears when tabbed into - allowing keyboard users to skip to the main content of a page.

Conflicts resolved with master 

Should be the first tab on the homepage (and on other pages following a page reload)


abijames:
@kprist please could you check the code on this PR and merge if ok? It adds a skip to content link to the header component & then add internal page links to support this. I've also improved headings, accessibility issues and minor typos on this PR.

package.json
```diff
@@ -210,5 +211,42 @@
     "sloc": "0.2.0",
     "webpack-dev-server": "^2.4.1",
     "webpack-vendor-chunk-plugin": "1.0.0"
+  },
```
kprist: @abijames @nk1g09 what is the section doing exactly? I noticed it creates a "pancake" folder in the root, do we need this?

I'm not sure how this will be deployed once the platform is build, specifically if the theming will be correct when deployed. Could we have a screenshot for reference ?

abijames:
@kprist this creates  a pop up menu that is shown when you first tab through the header ie focus is before the SlideWiki logo. So if you shift-tab back through the page through the header you will see the menu. The skip link should go to the #main link in the page

nk1g09:
@kprist the pancake folder comes from the skiplink installation. I don't believe it is necessary but I wasn't sure how to stop it. 

The styling has been modified and put into our css files as well so it is accessible.

PR #1153, opened 2019-02-22
Fix IE syntax error problem 

aoelen:
I have commented out some ES6 code that breaks SlideWiki in Internet Explorer. As far as I have seen, the code is not used by other components.


abijames:
Testing with first commit on 22/2/19. This fixed the IE11 issue but caused some CSS problems. I have been unable to test the later commits which intend to resolve this.

abijames:
Confirmed CSS on dropdowns working correctly on IE11.

PR #1157, opened 2019-03-28
Fix error when language is null

aoelen:
The caused for some slide decks a 504 error on page load, since an exception was thrown.

E.g. this deck cannot be loaded right now: https://slidewiki.org/deck/584 The page is rendered at the server-side, but because there is an exception, the page stops loading. After 60 seconds a 504 Gateway Time-out error will appear.


bwulff:
I bet that was some digging! Thanks a lot, Allard!

PR #1164, opened 2019-04-16
Fix alt-text quotes for images in slide editor 

### LINKED ISSUE #1162: Alt-text quotes not escaped
If you look at the image metadata in the store on [this presentation](https://slidewiki.org/deck/127208/alttext-tests), you'll see that US-style quote marks (") were saved in the underlying database, instead of html & code `&quot`.  That was achieved by entering the " symbol into the "caption" form while creating the figure.  This means that alt text cannot contain quotes entered as such.

aoelen:
Fixes #1162


bwulff:
Looking at the code @aoelen , thank you very much for making this the right way! DOM changes like these should always be done via API not by concatenating strings (although this is what I often do to).

PR #1175, opened 2019-06-17
bumped up versions of vulnerable packages

bwulff:


aoelen:
I've just reverted the version of webpack-dev-server which is not compatible with webpack 3. The upgrades diff package works fine!

bwulff:
Thanks for looking at it @aoelen !

PR #1183, opened 2019-07-10
Upgrade React to V16

aoelen:
Everything works as expected, apart from the `react-hot-loader`. Currently hot file loading has been disabled because the V1 version of `react-hot-loader` is not compatible with React 16.

Would be nice if somebody can take a look at upgrading to `react-hot-loader` to the newest version. Unfortunately I've been unsuccessful with this. As soon as hot loading is fixed, this can be merged.


aoelen:
Still working on this, managed to fix the react-hot-loader by upgrading this package and by upgrading webpack to V4. 

Extra testing is still needed. 

Needs to be fixed: 
* [ ] PaintModal (because in componentDidUpdate a DOM reference is used, which is not available during render)

aoelen:
New bug discovered: does not work in Internet Explorer (ES6 code in main.js, which causes a syntax error)

rezatavakoli:
We've checked the main functionality of the following list and they work well:

- Homepage
- Sign-in with user/pass
- Sign-in with gmail
- Create Account
- Reset Password
- My Deck List
- Select a Deck and view the deck home page
- Add/Edit/Delete Slide
- Go through the slides + presentation (next, previous, first, last, notes, go to deck page Overview, slides, report, downloads, share)
- Add/Edit/Delete Resources (slide/deck)
- Add/Reply/Delete Comments (slide/deck)
- Check History
- Add/Edit/Delete Tag (deck)
- Add/Edit Questions (deck)
- Fork Deck
- Upload/Remove/Use Gravatar in profile picture
- Edit Personal Data
- Search

PR #1249, opened 2020-04-15
1224 info relationships

### LINKED ISSUE #1224: Info and Relationships (WCAG 1.3.1h)
**Request:** The intent of this Success Criterion is to ensure that information and relationships that are implied by visual or auditory formatting are preserved when the presentation format changes. For example, the presentation format changes when the content is read by a screen reader or when a user style sheet is substituted for the style sheet provided by the author.

**Issues:**

1. https://slidewiki.org/search?keywords=accessibility&sort=score

- The form fields within the "Advanced options" are not linked to their labels above. The corresponding form elements should be labeled with a label-element and be linked to the label above via aria-labeledby.

2. https://slidewiki.org/addDeck

- Labeling of the form fields "Wählen Sie das Thema" and "Wählen Sie die Tags" is not correctly linked to the associated input fields. The label and form field can be linked using aria-labeledby.

rezatavakoli:
The following changes have been applied on the Search page:
Under advanced options:
- For language-part: aria-labelledby has been added.
- For subject, owners, and tags parts aria-labelled passes through props of the related components

The following changes have been applied on Add deck page
- For the theme part, id has been set in order to solve the accessibility problem.
- For tags part, aria-labelled passes through related component

Fixes #1224


aoelen:
Fixes #1224

PR #1291, opened 2022-02-24
chore: upgrade node version Dockerfile

### LINKED ISSUE #1290: Data not loading: Error: certificate has expired
Data isn't loading current, this error is displayed `Error: certificate has expired`

Upgrading the node version seems to fix the issue

aoelen:
Fixes #1290


bwulff:
Did we check if we loose functionality when using the node image directly instead of the slidewiki-runtime image?

